PR #21, opened 2018-05-23
PLATOPS-1308 - Show the Slack Notification Channel in Catalogue

jakobgrunig:


app/uk/gov/hmrc/cataloguefrontend/connector/TeamsAndRepositoriesConnector.scala
```diff
@@ -126,26 +126,16 @@ class TeamsAndRepositoriesConnector @Inject()(
...
     http
-      .GET[HttpResponse](url)
-      .map { response =>
-        response.status match {
-          case 404 => None
-          case 200 => Some(toJson[Team](response))
-        }
-      }
+      .GET[Option[Team]](url)
       .recover {
```
dabd: Maybe we can get rid of this logging since http-verbs already does it https://github.com/hmrc/http-verbs/blob/master/src/main/scala/uk/gov/hmrc/http/logging/ConnectionTracing.scala#L29?

PR #23, opened 2018-08-09
[WIP] Feature/play 26

alexflav23:
- Replace `AbstractController` with `MessagesAbstractController` to deal with changes in I8n support in play 2.6
- Replace `ControllerComponents` with `MessagesControllerComponents` to deal with changes in I8n support in play 2.6
- Inject `ViewMessages` that rely on configuration instead of using a Scala `object ViewMessages`.
- Change all templates to use an injected version of `ViewMessages` passed through from the instantiating controller instead of using the singleton import.
- Rename inner class `Environment` to `TargetEnvironment` to prevent collision with Play.
- Replace `ServicesConfig` inclusion to work with injection instead of inheritance.
- Manually specify an `implicit lazy val defaultLang: Lang = Lang(java.util.Locale.getDefault)` in places where `Messages` are accessed . It seems inside controllers and compile scope this is solved by Play `I8nSupport`, in `Test` scope we need to manually add this to places.
- Replace `OneAppPerSuite` and `OneAppPerTest` with `GuiceOneAppPerSuite` and `GuiceOneAppPerTest`
- Replace import for `GuiceOneAppPerSuite` and `GuiceOneAppPerTest` to `import org.scalatestplus.play.guice._` instead of `import org.scalatestplus.play._`.

h4. Suggested improvements.

- Remove the `RunMode` abstraction layer from `play-bootstrap`. It provides something we already get natively from `Mode` inside play, and if we are required to add a method that returns extra metadata depending on the mode and `Configuration`, we can implement that using an `implicit class`, therefore removing the coupling with runtime injection.

### DIFF
--- a/app/uk/gov/hmrc/cataloguefrontend/AuthController.scala
+++ b/app/uk/gov/hmrc/cataloguefrontend/AuthController.scala
@@ -34,8 +34,9 @@ import scala.concurrent.Future
 @Singleton
 class AuthController @Inject()(
   authService: AuthService,
-  configuration: Configuration
-) extends FrontendController {
+  configuration: Configuration,
+  mcc: MessagesControllerComponents
+) extends MessagesAbstractController(mcc) with FrontendController {
 
   import AuthController.signinForm
 
--- a/app/uk/gov/hmrc/cataloguefrontend/AuthController.scala
+++ b/app/uk/gov/hmrc/cataloguefrontend/AuthController.scala
@@ -20,8 +20,8 @@ import javax.inject.{Inject, Singleton}
 import play.api.Configuration
 import play.api.data.Form
 import play.api.data.Forms._
-import play.api.i18n.{I18nSupport, Messages, MessagesApi}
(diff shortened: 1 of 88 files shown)


app/uk/gov/hmrc/cataloguefrontend/AuthController.scala
```diff
@@ -32,22 +32,21 @@ import views.html.sign_in
...
   import AuthController.signinForm
 
-  private val selfServiceUrl = {
-    val key = "self-service-url"
-    configuration.getString(key).getOrElse(throw new Exception(s"Expected to find $key"))
-  }
+  private[this] val selfServiceUrl = configuration.get[String]("self-service-url")
 
-  val showSignInPage = Action { implicit request =>
+  val showSignInPage: Action[AnyContent] = Action { implicit request =>
```
konradwudkowski: We usually don't add `: Action[AnyContent]`, it's not wrong defnitely and feel free to make a decision to keep it or not

PR #24, opened 2018-08-21
Feature/play 26

alexflav23:
- Replace `AbstractController` with `MessagesAbstractController` to deal with changes in I8n support in play 2.6
- Replace `ControllerComponents` with `MessagesControllerComponents` to deal with changes in I8n support in play 2.6
- Inject `ViewMessages` that rely on configuration instead of using a Scala `object ViewMessages`.
- Change all templates to use an injected version of `ViewMessages` passed through from the instantiating controller instead of using the singleton import.
- Rename inner class `Environment` to `TargetEnvironment` to prevent collision with Play.
- Replace `ServicesConfig` inclusion to work with injection instead of inheritance.
- Manually specify an `implicit lazy val defaultLang: Lang = Lang(java.util.Locale.getDefault)` in places where `Messages` are accessed . It seems inside controllers and compile scope this is solved by Play `I8nSupport`, in `Test` scope we need to manually add this to places.
- Replace `OneAppPerSuite` and `OneAppPerTest` with `GuiceOneAppPerSuite` and `GuiceOneAppPerTest`
- Replace import for `GuiceOneAppPerSuite` and `GuiceOneAppPerTest` to `import org.scalatestplus.play.guice._` instead of `import org.scalatestplus.play._`.

h4. Suggested improvements.

- Remove the `RunMode` abstraction layer from `play-bootstrap`. It provides something we already get natively from `Mode` inside play, and if we are required to add a method that returns extra metadata depending on the mode and `Configuration`, we can implement that using an `implicit class`, therefore removing the coupling with runtime injection.


app/uk/gov/hmrc/cataloguefrontend/AuthController.scala
```diff
@@ -32,22 +32,21 @@ import views.html.sign_in
...
 
 @Singleton
-class AuthController @Inject()(val messagesApi: MessagesApi, authService: AuthService, configuration: Configuration)
-    extends FrontendController
-    with I18nSupport {
+class AuthController @Inject()(
+  authService: AuthService,
+  configuration: Configuration,
+  mcc: MessagesControllerComponents
+) extends MessagesAbstractController(mcc) with FrontendController {
```
konradwudkowski: alternative is to do:
```
...
  val controllerComponents: MessagesControllerComponents
) extends FrontendController {
```
We should find the best way here for teams to use as this will be one of the most common changes they'll need to do

app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
```diff
@@ -63,38 +63,37 @@ class CatalogueController @Inject()(
   environment: api.Environment,
   verifySignInStatus: VerifySignInStatus,
   umpAuthenticated: UmpAuthenticated,
-  override val runModeConfiguration: Configuration,
-  val messagesApi: MessagesApi)
-    extends FrontendController
-    with UserManagementPortalLink
-    with I18nSupport {
+  val serviceConfig: ServicesConfig,
```
konradwudkowski: service**s**Config perhaps to keep it consistent

konradwudkowski:
Not sure about these two comments as we never had AbstractController or ControllerComponents before:

- Replace AbstractController with MessagesAbstractController to deal with changes in I8n support in play 2.6
- Replace ControllerComponents with MessagesControllerComponents to deal with changes in I8n support in play 2.6

app/uk/gov/hmrc/cataloguefrontend/FutureHelpers.scala
```diff
@@ -16,87 +16,51 @@
...
-        fn
-      })
-
-    def filter[B](fn: A => Boolean)(implicit ec: ExecutionContext): Future[Iterable[A]] =
-      futureList.map(_.filter(fn))
+    def apply[A](listFuture: Iterable[Future[A]]): Future[Iterable[A]] = Future.sequence(listFuture)
   }
 
-  def continueOnError[A](f: Future[A]) =
+  def continueOnError[A](f: Future[A]): Future[Try[A] with Product with Serializable] =
```
konradwudkowski: Just `Future[Try[A]]` is probably enough as return type here

app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
jachro: What about making the `UserManagementPortalLink` a class with `ServicesConfig` injected into it (currently we need to inject `ServicesConfig` just because of that trait). Then, the portal link instance can be injected into the `CatalogueController` as a collaborator. It doesn't look very right to me that the `UserManagementPortalLink`'s api enriches the controller's api and you cannot say the `CatalogueController` is a `UserManagementPortalLink`.

app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
```diff
@@ -227,16 +227,17 @@ class CatalogueController @Inject()(
               TeamChartData.deploymentStability(team.name, teamIndicators.map(_.stability)),
               umpMyTeamsPageUrl(team.name),
               leakDetectionService.teamHasLeaks(team, reposWithLeaks),
-              leakDetectionService.hasLeaks(reposWithLeaks)
+              leakDetectionService.hasLeaks(reposWithLeaks),
```
jachro: What do you think about changing the `service_info` view by creating an @Inject annotated args list and moving some of the existing args into that? I'm guessing it could de-clutter the controller's args and reduce the number of args we have to pass to `service_info`.

app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
```diff
@@ -293,39 +294,40 @@ class CatalogueController @Inject()(
               ServiceChartData.deploymentStability(repositoryDetails.name, maybeDataPoints.map(_.stability)),
               repositoryDetails.createdAt,
               deploymentsByEnvironmentName,
-              urlIfLeaksFound
+              urlIfLeaksFound,
```
jachro: Maybe we could do the same to `service_info` as I mentioned in the comment few lines above for the `team_info`?

app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
```diff
@@ -293,39 +294,40 @@ class CatalogueController @Inject()(
...
+  def library(name: String): Action[AnyContent] = Action.async { implicit request =>
     for {
       library           <- teamsAndRepositoriesConnector.repositoryDetails(name)
       mayBeDependencies <- serviceDependencyConnector.getDependencies(name)
       urlIfLeaksFound   <- leakDetectionService.urlIfLeaksFound(name)
     } yield
       library match {
         case Some(s) if s.repoType == RepoType.Library =>
-          Ok(library_info(s, mayBeDependencies, urlIfLeaksFound))
+          Ok(library_info(s, mayBeDependencies, urlIfLeaksFound, viewMessages))
```
jachro: ditto and actually in few other places below.

app/uk/gov/hmrc/cataloguefrontend/DependencyReportController.scala
```diff
@@ -48,18 +48,20 @@ case class DependencyReport(
...
-    with UserManagementPortalLink {
-
-  override protected def mode = environment.mode
+  serviceDependencyConnector: ServiceDependenciesConnector,
+  val serviceConfig: ServicesConfig,
+  mcc: MessagesControllerComponents
+) extends MessagesAbstractController(mcc) with FrontendController with UserManagementPortalLink {
 
-  implicit val drFormat = Json.format[DependencyReport]
+  implicit val drFormat: OFormat[DependencyReport] = Json.format[DependencyReport]
```
jachro: Can that be private?

app/uk/gov/hmrc/cataloguefrontend/actions/UmpAuthenticated.scala
```diff
@@ -45,4 +48,7 @@ class UmpAuthenticated @Inject()(userManagementAuthConnector: UserManagementAuth
     }
   }
 
+  override def parser: BodyParser[AnyContent] = cc.parsers.defaultBodyParser
+
+  override protected def executionContext: ExecutionContext = cc.executionContext
```
jachro: Here comes this a bit awkward and possibly confusing thing for people (effectively it's all about the problem of whether to use MdcLoggingExecutionContext or Slf4jLogger's Markers). Because of ActionBuilder we are forced to provide an execution context and we have to take it from ControllerComponents and at the same time, in the `invokeBlock` we use the implicit conversion from HeaderCarrier => MdcLoggingExecutionContext. Personally, I don't really like this ambiguity and I'm not sure if people won't get lost and understand what's really going on. What about we spend some time and think thoroughly how to tackle this issue.

app/uk/gov/hmrc/cataloguefrontend/actions/VerifySignInStatus.scala
```diff
@@ -47,4 +49,7 @@ class VerifySignInStatus @Inject()(userManagementAuthConnector: UserManagementAu
     }
   }
 
+  override def parser: BodyParser[AnyContent] = cc.parsers.anyContent
+
+  override protected def executionContext: ExecutionContext = cc.executionContext
```
jachro: Ditto

app/uk/gov/hmrc/cataloguefrontend/connector/IndicatorsConnector.scala
```diff
@@ -87,12 +87,10 @@ case class MeasureResult(median: Int)
 @Singleton
 class IndicatorsConnector @Inject()(
   http: HttpClient,
-  override val runModeConfiguration: Configuration,
-  environment: Environment)
-    extends ServicesConfig {
-  def indicatorsBaseUrl = baseUrl("indicators") + "/api/indicators"
-
-  override protected def mode = environment.mode
+  environment: Environment,
```
jachro: It looks the `environment` arg is not used.

app/uk/gov/hmrc/cataloguefrontend/connector/LeakDetectionConnector.scala
```diff
@@ -17,24 +17,23 @@
...
-import uk.gov.hmrc.play.config.ServicesConfig
 import uk.gov.hmrc.play.http.logging.MdcLoggingExecutionContext._
 
+import scala.concurrent.Future
+
 @Singleton
 class LeakDetectionConnector @Inject()(
   http: HttpClient,
-  override val runModeConfiguration: Configuration,
-  environment: Environment)
```
jachro: It looks the environment arg is not used.

app/uk/gov/hmrc/cataloguefrontend/connector/ServiceDependenciesConnector.scala
```diff
@@ -33,26 +33,23 @@ package uk.gov.hmrc.cataloguefrontend.connector
...
 import scala.concurrent.Future
 @Singleton
 class ServiceDependenciesConnector @Inject()(
   http: HttpClient,
-  override val runModeConfiguration: Configuration,
-  environment: PlayEnvironment)
-    extends ServicesConfig {
-
-  def servicesDependenciesBaseUrl: String = baseUrl("service-dependencies") + "/api"
+  environment: PlayEnvironment,
```
jachro: It looks the environment arg is not used.

app/uk/gov/hmrc/cataloguefrontend/connector/ServiceDeploymentsConnector.scala
```diff
@@ -80,14 +80,12 @@ object ServiceDeploymentInformation {
 @Singleton
 class ServiceDeploymentsConnector @Inject()(
   http: HttpClient,
-  override val runModeConfiguration: Configuration,
-  environment: PlayEnvironment)
-    extends ServicesConfig {
+  environment: PlayEnvironment,
```
jachro: It looks the environment arg is not used.

app/uk/gov/hmrc/cataloguefrontend/service/LeakDetectionService.scala
```diff
@@ -36,26 +36,12 @@ class LeakDetectionService @Inject()(leakDetectionConnector: LeakDetectionConnec
...
-      .getBoolean(key)
-      .getOrElse(
-        throw new Exception(s"Failed reading from config, expected to find: $key")
-      )
-  }
+  private val ldsIntegrationEnabled: Boolean = configuration.get[Boolean]("lds.integrationEnabled")
 
   val repositoriesToIgnore: List[String] =
-    configuration.getStringList("lds.noWarningsOn").fold(List.empty[String])(_.asScala.toList)
+    configuration.underlying.getStringList("lds.noWarningsOn").asScala.toList
```
jachro: Maybe a good candidate to add to the Bootstrap an implicit ConfigLoader for List[String]?

app/views/digital_service_info.scala.html
```diff
@@ -14,16 +14,17 @@
...
-@import uk.gov.hmrc.cataloguefrontend.ViewMessages._
+@import uk.gov.hmrc.cataloguefrontend.{DigitalServiceDetails, ViewMessages}
+@import uk.gov.hmrc.cataloguefrontend.connector.RepoType
+@import uk.gov.hmrc.cataloguefrontend.connector.RepoType.RepoType
 @import uk.gov.hmrc.cataloguefrontend.DisplayableTeamMember
 @import uk.gov.hmrc.cataloguefrontend.actions.UmpVerifiedRequest
 
-@(timestampedRepos: DigitalServiceDetails,
-  digitalServiceOwner: Option[DisplayableTeamMember]
+@(
```
jachro: My subjective preference when talking about template refactoring would be to use injection for things like ViewMessages. The benefit of that would be that it takes that necessity of injecting things needed by a template to a controller which effectively does not need it. I believe such approach makes things cleaner and easier to reason about. What do you think?

test/uk/gov/hmrc/cataloguefrontend/AuthControllerSpec.scala
```diff
@@ -97,22 +100,24 @@ class AuthControllerSpec extends WordSpec with Matchers with OneAppPerSuite with
     "redirect to landing page and clear session" in new Setup {
       val requestWithUmpData = FakeRequest()
 
+      val sessionInjector: SessionCookieBaker = app.injector.instanceOf[SessionCookieBaker]
       val result = controller.signOut(requestWithUmpData)
 
-      val setCookie = Cookies.decodeSetCookieHeader(headers(result).get(SET_COOKIE).get).head
+      //val setCookie: Cookie = Cookies.decodeSetCookieHeader(headers(result).apply(SET_COOKIE)).headOption.value
```
jachro: Would be nice to fix this test in a way we think people should do it.

test/uk/gov/hmrc/cataloguefrontend/CatalogueControllerSpec.scala
```diff
@@ -140,10 +142,11 @@ class CatalogueControllerSpec extends WordSpec with MockitoSugar {
...
-    val configuration      = mock[Configuration]
+    val configuration      = mock[ServicesConfig]
+    val mcc = mock[MessagesControllerComponents]
 
-    when(configuration.getString("microservice.services.user-management.profileBaseUrl"))
-      .thenReturn(Some("profile-base-url"))
+    when(configuration.getConfString("microservice.services.user-management.profileBaseUrl", ""))
+      .thenReturn("profile-base-url")
 
     val controller = new CatalogueController(
```
jachro: There is a general question of how we would like to use templates: inject them or reference them through object defs? If we'd be injecting them, probably we can mock them in specs' setups and stop using GuiceOneAppPerSuite and what maybe even more important, controllers tests will be purely verifying logic coded in the endpoints not trying to find something suitable in a template to use assertions like: `html includes something` which doesn't seems to be very robust and easy to understand. I know this approach comes with a cost of doing a bit more changes to the existing specs but maybe we could suggest it to the teams as an option for existing services and a prefered approach for new services?

app/uk/gov/hmrc/cataloguefrontend/UserManagementPortalLink.scala
```diff
@@ -16,22 +16,27 @@
 
 package uk.gov.hmrc.cataloguefrontend
 
-import uk.gov.hmrc.play.config.ServicesConfig
+import uk.gov.hmrc.play.bootstrap.config.ServicesConfig
+
+
+trait UserManagementPortalLink {
```
konradwudkowski: Let's convert to a class that we inject where needed

app/uk/gov/hmrc/cataloguefrontend/actions/UmpAuthenticated.scala
jachro: To be addressed as part of https://jira.tools.tax.service.gov.uk/browse/PLATOPS-1449

PR #29, opened 2018-11-27
BDOG-37 Fixed issue with position: sticky in safari

christopherjturner:


fenallen:
This is now rather old and no longer compiles due to dependency issues.
Closing this pr and applying the changes to a new branch is simpler than fixing this.

PR #75, opened 2019-07-12
Bdog 238

colin-lamed:
Display shuttering status, and add wizard to shutter/unshutter services.

### DIFF
--- a/app/views/shuttering/shutterService/Page2.scala.html
+++ b/app/views/shuttering/shutterService/Page2.scala.html
@@ -52,7 +52,9 @@
             <tr>
               <td>
                 <ul style="list-style: disc;">
-                  <li>@shutter.serviceName</li>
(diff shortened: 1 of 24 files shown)


app/views/shuttering/shutterService/Page2.scala.html
```diff
@@ -0,0 +1,95 @@
...
+  </header>
+
+  <div id="service-list">
+
+    <div class="board">
+      <div class="board__body">
+        @helper.form(
+            action =  uk.gov.hmrc.cataloguefrontend.shuttering.routes.ShutterServiceController.step2Post
+          , 'id    -> "shutter-service-form"
+          , 'style -> "padding-left: 20px; padding-right: 20px"
```
christopherjturner: any value in moving the style tags into css classes? that said, the main .css file is pretty huge so perhaps not

PR #77, opened 2019-07-23
Bdog 309

colin-lamed:
Shuttering service workflow:
- Add Outage page feedback
- Improve service selection with groups.

### DIFF
--- a/app/uk/gov/hmrc/cataloguefrontend/shuttering/ShutterConnector.scala
+++ b/app/uk/gov/hmrc/cataloguefrontend/shuttering/ShutterConnector.scala
@@ -31,8 +31,9 @@ class ShutterConnector @Inject()(
   , serviceConfig: ServicesConfig
   )(implicit val ec: ExecutionContext){
 
-  private val urlStates: String = s"${serviceConfig.baseUrl("shutter-api")}/shutter-api/states"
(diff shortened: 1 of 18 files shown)


app/uk/gov/hmrc/cataloguefrontend/shuttering/ShutterConnector.scala
```diff
@@ -88,4 +93,15 @@ class ShutterConnector @Inject()(
...
+
+
+  /**
+    * GET
+    * /shutter-api/outage-pages/{appName}/{environment}
+    * Retrieves the current shutter state for the given application in the given environment
+    */
+  def outagePageByAppAndEnv(appName: String, env: Environment)(implicit hc: HeaderCarrier): Future[Option[OutagePage]] = {
+    implicit val ssf = OutagePage.reads
+    http.GET[Option[OutagePage]](s"$urlOutagePages/$appName/${env.asString}")
```
thomas-vogel444: Are you sure this will give you a None if the outage page isn't there? Won't it fail the future with a 4xxException?

colin-lamed: Yup https://github.com/hmrc/http-verbs/blob/master/src/main/scala/uk/gov/hmrc/http/HttpReads.scala#L44

PR #80, opened 2019-07-30
BDOG-302 added jdk vendor information

christopherjturner:


app/views/JdkVersionPage.scala.html
```diff
@@ -36,7 +36,17 @@ <h1 id="jdk-header">JDK Versions: @selected.asString</h1>
         <script type="text/javascript" src="https://www.gstatic.com/charts/loader.js"></script>
     </header>
 
-    <div id="chart_div"></div>
+    <div class="row">
```
chotai: Do we want to show Kind as well?

PR #83, opened 2019-08-07
BDOG-302 added vendor icon, better sorting, jdk kind data

christopherjturner:


app/uk/gov/hmrc/cataloguefrontend/JDKVersionController.scala
```diff
@@ -41,16 +41,18 @@ class JDKVersionController @Inject()(
     for {
         flag        <- Future.successful(SlugInfoFlag.parse(flag.toLowerCase).getOrElse(SlugInfoFlag.Latest))
         jdkVersions <- dependenciesService.getJDKVersions(flag)
-    } yield Ok(jdkPage(jdkVersions, SlugInfoFlag.values, flag))
+    } yield Ok(jdkPage(jdkVersions.sortBy(byJDKVersion), SlugInfoFlag.values, flag))
   }
 
   def compareAllEnvironments() = Action.async { implicit request =>
-    implicit val r = JDKUsageByEnvFormat.jdkUsageByEnvFormat
+    //implicit val r = JDKUsageByEnvFormat.jdkUsageByEnvFormat
```
bennetimo: Is this supposed to be commented?

christopherjturner: should probably be deleted rather than commented. endpoints return html rather than json so formatters are not required

bennetimo: Ah ok cool, makes sense

PR #91, opened 2019-08-20
Bdog 366

bennetimo:
1. Add page at /frontend-route-warnings/:env/:service for showing all of the frontend-route warnings for a service retrieved from the `shutter-api`
1. Split the `/shuttering/2` page in the shutter service flow in two, `/shuttering/2a` and `/shuttering/2b`. The new `2a` page displays any frontend route warnings and requires the user to confirm they have acknowledged them before they can continue. The flow then continues to `2b` which is the outage-page step as before.
1. Page `2a` is skipped if there are no warnings for any of the services being shuttered (or if being unshuttered)

### DIFF
--- a/app/uk/gov/hmrc/cataloguefrontend/shuttering/ShutterController.scala
+++ b/app/uk/gov/hmrc/cataloguefrontend/shuttering/ShutterController.scala
@@ -25,6 +25,7 @@ import views.html.shuttering._
 
 import scala.concurrent.ExecutionContext
 import scala.util.control.NonFatal
+import cats.implicits._
 
 @Singleton
 class ShutterController @Inject()(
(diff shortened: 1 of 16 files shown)


app/uk/gov/hmrc/cataloguefrontend/shuttering/ShutterController.scala
```diff
@@ -23,29 +23,47 @@ import uk.gov.hmrc.cataloguefrontend.actions.VerifySignInStatus
...
+                             Logger.error(s"Could not retrieve currentState: ${ex.getMessage}", ex)
+                             Seq.empty
+                         }
+        page = shutterStatePage(currentState, env, request.isSignedIn)
+      } yield Ok(page)
+    }
+
+  def frontendRouteWarnings(envParam: String, serviceName: String): Action[AnyContent] =
+    Action.async { implicit request =>
+      val env = Environment.parse(envParam).getOrElse(Environment.Production)
```
thomas-vogel444: Why defaulting to production instead of returning a 404?

PR #131, opened 2019-11-12
BDOG-448: add Shutter Events view

nigelhp:


app/uk/gov/hmrc/cataloguefrontend/shuttering/ShutterConnector.scala
```diff
@@ -117,3 +124,16 @@ class ShutterConnector @Inject()(
       .GET[Seq[FrontendRouteWarning]](s"${urlFrontendRouteWarnings(env)}/${encodePathParam(serviceName)}")
   }
 }
+
+object ShutterConnector {
+  case class ShutterEventsFilter(environment: Environment, serviceName: Option[String])
+
+  object ShutterEventsFilter {
+    private [shuttering] def asQuery(filter: ShutterEventsFilter): String =
+      asQueryParam(key = "data.environment", value = filter.environment.asString) +
```
colin-lamed: maybe move `&` out of queryParam and into join e.g. `(Seq(queryParam("data.environment", environment) + serviceName.map(queryParam("data.serviceName", _)).join("&")`
This will move the initial `&` into the url (where we have context of if a previous query param has been used)

app/uk/gov/hmrc/cataloguefrontend/shuttering/ShutterConnector.scala
```diff
@@ -93,6 +94,12 @@ class ShutterConnector @Inject()(
         s"$urlEvents?type=${encodeQueryParam(EventType.ShutterStateChange.asString)}&namedFilter=latestByServiceName&data.environment=${encodeQueryParam(env.asString)}&data.shutterType=${encodeQueryParam(st.asString)}")
       .map(_.flatMap(_.toShutterStateChangeEvent))
 
+  def shutterEventsByTimestampDesc(filter: ShutterEventsFilter)(implicit reads: HttpReads[Seq[ShutterEvent]], hc: HeaderCarrier): Future[Seq[ShutterStateChangeEvent]] = {
```
colin-lamed: The `implicit HttpReads[ShutterEvent]` should be provided by this function (internal import), not a parameter, since it is part of the Api - not flexible.
In fact it is already in scope (line 44)

app/uk/gov/hmrc/cataloguefrontend/shuttering/model.scala
```diff
@@ -124,14 +141,22 @@ object ShutterStatusValue {
   }
 }
 
-sealed trait ShutterStatus { def value: ShutterStatusValue }
+sealed trait ShutterStatus {
+  def value: ShutterStatusValue
+  def reason: Option[String]
```
colin-lamed: Whether reason was present or not was previously encoded in the type. Has this been added just to help the UI?
I'd be tempted to add a function for this to `object ShutterStatus`, or even `ShutterEventPage` e.g. `def optReason(ss: ShutterStatus): Option[String]` if required

app/uk/gov/hmrc/cataloguefrontend/shuttering/ShutterConnector.scala
nigelhp: Agree that seems cleaner.

app/uk/gov/hmrc/cataloguefrontend/shuttering/ShutterConnector.scala
nigelhp: I had reluctantly exposed the ```HttpReads``` in order to put a unit test together for the connector which uses a mock ```Http```.  I have now reworked this using an ```any``` matcher, and so have been able to hide this again.

app/uk/gov/hmrc/cataloguefrontend/shuttering/model.scala
nigelhp: This is purely for the UI, and so I agree that it doesn't really belong in the model.
Have added a ```ShutterEventsView``` object as a helper object for the view to contain this function.  Although a simple function, I didn't want to add this into the template itself.

PR #138, opened 2019-11-21
BDOG-447: Filtering, mainly by App Name

cjwebb:
Profile filter can be done via the query params, but there's another
ticket somewhere to do profiles properly so haven't completed that bit
yet.
The application-name filtering is only via JavaScript, as there are a
few UX behaviours that feel weird with the current way we filter
server-side. I will spend a bit of time trying to form a model for how
we do this nicely...

Screenshot available on the JIRA.

### DIFF
--- a/app/views/WhatsRunningWherePage.scala.html
+++ b/app/views/WhatsRunningWherePage.scala.html
@@ -16,10 +16,11 @@
 
 @import uk.gov.hmrc.cataloguefrontend.ViewMessages
 @import uk.gov.hmrc.cataloguefrontend.whatsrunningwhere.{Environment, WhatsRunningWhere}
+@import uk.gov.hmrc.cataloguefrontend.routes
 
 @this(viewMessages: ViewMessages)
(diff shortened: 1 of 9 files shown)


app/views/WhatsRunningWherePage.scala.html
```diff
@@ -58,7 +58,7 @@ <h1>What's Running Where?</h1>
             <tbody class="list">
             @whatsRunning.zipWithIndex.map { case (wrw, i) =>
             <tr role="row" id="row@{i}_name">
-                <td class="application-name">@{wrw.applicationName.asString}</td>
+                <td class="application-name"><a href="/service/@{wrw.applicationName.asString}">@{wrw.applicationName.asString}</a></td>
```
colin-lamed: Can you use reverse routes, rather than hard-coding the url (it will then be validated in compile time)

cjwebb: Yes, my bad. Have pushed again with a resolution to this.

PR #143, opened 2019-12-04
BDOG-446: service page - environments as tabs with version-specific dependencies

nigelhp:
Add a tabbed section to the service page, and move library dependencies & service urls (such as Kibana / Grafana) into the tab - so that the information displayed relates to the selected environment / deployed version of the service in that environment.

Note that the platform dependencies displayed on the "latest" tab are sourced from master (as at present) i.e. they are not sourced from a slug, whereas the dependencies displayed on all other tabs are pulled from the slug relating to the version of the service deployed in that environment.

### DIFF
--- a/public/catalogue-frontend.css
+++ b/public/catalogue-frontend.css
@@ -12283,7 +12283,6 @@ table.shutter-events {
 
 #environmentTabs li a .board .board__heading {
   border: none;
-  padding: 6px;
   padding-top: 2px;
   padding-bottom: 0px;
 }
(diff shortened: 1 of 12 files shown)


public/catalogue-frontend.css
```diff
@@ -12286,4 +12269,60 @@ table.shutter-events {
...
+  background: inherit;
+}
+
+.environment-panes {
+  padding: 16px;
+  padding-bottom: 0px;
+  margin-bottom: 16px;
+  border: 1px solid #ddd;
+  border-top: none;
+}
```
cjwebb: I would like to suggest these changes. I think the CSS needs more work than this though, at some point in time.

```
#environmentTabs li a .board .board__heading {
   border: none;
-  padding: 6px;
   padding-top: 2px;
-  padding-bottom: 0px;
 }

-.nav-tabs > li.active > a > .board > .board__heading {
+.nav-tabs > li > a > .board > .board__heading {
   background: inherit;
 }

+.environment-panes .board {
+  border: none;
+}

```

PR #148, opened 2019-12-17
Bdog 525 test dependencies

nigelhp:
Assume that library dependencies derived from parsing build files on master that are not known to the appropriate "latest" tagged "slug-info" represent "test-only" library dependencies.  Examples are typically _reactivemongo-test_ and _hmrctest_.  Add such dependencies to what was previously called the "Build Tools" section under the heading "Test libraries", and rename the section to "Latest Build" (as it is no longer just about "tools").

Sort the display of dependencies by name - they are currently displayed in the order retrieved from the database.

### DIFF
--- a/app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
+++ b/app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
@@ -28,6 +28,7 @@ import play.api.mvc._
 import uk.gov.hmrc.cataloguefrontend.DisplayableTeamMember._
 import uk.gov.hmrc.cataloguefrontend.actions.{UmpAuthActionBuilder, VerifySignInStatus}
 import uk.gov.hmrc.cataloguefrontend.connector.RepoType.Library
+import uk.gov.hmrc.cataloguefrontend.connector.SlugInfoFlag.Latest
 import uk.gov.hmrc.cataloguefrontend.connector.UserManagementConnector.UMPError
 import uk.gov.hmrc.cataloguefrontend.connector._
 import uk.gov.hmrc.cataloguefrontend.events._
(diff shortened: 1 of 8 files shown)


app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
```diff
@@ -305,18 +306,11 @@ class CatalogueController @Inject()(
...
-           * Slugs do not contain build related information however.  So 'Build Tools' continues to display information
-           * parsed from master / GitHub.
-           */
           Ok(
             serviceInfoPage(
               repositoryDetails.copy(environments = optDeployedEnvironments, jenkinsURL = jenkinsLink),
-              optMasterDependencies.map(_.copy(libraryDependencies = librariesOfLatestSlug)),
-              librariesBySlugVersion,
+              optMasterDependencies,
+              librariesBySlugVersion + (Latest.asString -> librariesOfLatestSlug),
```
colin-lamed: if you just import `SlugInfoFlag` above, we get a namespace to give context (`SlugInfoFlag.Latest`)

colin-lamed: Can `serviceInfoPage` take `SlugInfoFlag` as key rather than String?
(Similarly deploymentsByEnvironmentName?)

nigelhp: Most of the slugs are being loaded by an explicit deployment version number.  However, we now also need to load "latest" - which uses a different API.  That API uses `SlugInfoFlag` - but we only care about one instance of those flags - `Latest`.  Essentially our key should be an explicit `Version` or some indicator of "latest version".

At the service level I introduced a wrapper algebraic type to capture this.
However, this hasn't been brought into the UI - because the UI seemed to be mostly String-typed - and it felt like it would be at odds with the existing code.

I didn't namespace `SlugInfoFlag.Latest` as this code doesn't really reference `SlugInfoFlag`.  We are just using the constant String "latest" to distinguish this from something like "1.2.3" - it could have been any String as long as it does not match a real version - and it just so happens that the constant that seemed most sensible was already defined.

I agree that this is only possible because the key is a String, and that it does feel "dirty".
The page currently treats "latest" specially anyway.  The data was added in this way simply to avoid adding another parameter - but it could just as easily have been passed separately - leaving `librariesBySlugVersion` to just contain the explicitly versioned slugs.  I'm happy to do that, in effect keeping the "flag" world separate from the "version" world.

I haven't changed `deploymentsByEnvironmentName`, but that would be possible.  I suspect that one of the issues here is that this information is sourced from _deployments-service_ which doesn't know anything about `SlugInfoFlag` (which is a _dependencies-service_ concept).  As you can see from this function in `CatalogueController` there is a lot of data being pulled from disparate sources here - and then much of the assembly is being done in the page itself.  There is quite a lot of complexity here, and it would be nice to spend some time abstracting over the various services and building a model for the view to simply display.

nigelhp: Rationale as to why I wanted the focus to be on `Latest`, and why the fact it comes from `SlugInfoFlag` isn't actually of interest here below.

PR #156, opened 2020-01-13
BDOG-453 Add TeamName to model

colin-lamed:


test.txt
```diff
@@ -0,0 +1 @@
+[0m[[0minfo[0m] [0m* 2.11.11[0m
```
bennetimo: Is this supposed to be here?

PR #162, opened 2020-01-20
BDOG-521: Show ECS releases in Catalogue

cjwebb:
Basic functionality to start with: a copy of WhatsRunningWhere, but just for ECS.

Lots more work, and refactoring, to come.

### DIFF
--- a/app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/model.scala
+++ b/app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/model.scala
@@ -16,7 +16,7 @@
 
 package uk.gov.hmrc.cataloguefrontend.whatsrunningwhere
 
-import java.time.LocalDateTime
+import java.time.{LocalDateTime, ZoneOffset}
 
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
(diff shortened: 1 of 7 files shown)


app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/model.scala
```diff
@@ -78,14 +79,43 @@ object JsonCodecs {
...
+    ~ (__ \ "deploymentEvents").format[Seq[DeploymentEvent]]
+    ~ (__ \ "lastCompleted").formatNullable[DeploymentEvent]
+    )(ServiceDeployments.apply, unlift(ServiceDeployments.unapply))
   }
 }
 
 case class TimeSeen(time: LocalDateTime)
 
+object TimeSeen {
+  implicit val localDateOrdering: Ordering[LocalDateTime] = Ordering.by(_.toEpochSecond(ZoneOffset.UTC))
```
colin-lamed: `localDateOrdering` should only be implicit in the scope of creating `timeSeenOrdering` - TimeSeen should not be introducing an implicit LocalDateTime.

cjwebb: Have pushed a small alteration to address this.

PR #165, opened 2020-01-22
Bdog 551 Unify Environment model

colin-lamed:


app/uk/gov/hmrc/cataloguefrontend/model/Environment.scala
```diff
@@ -0,0 +1,61 @@
...
+  case object ExternalTest    extends Environment { val asString = "externaltest" }
+  case object Staging         extends Environment { val asString = "staging"      }
+  case object QA              extends Environment { val asString = "qa"           }
+  case object Integration     extends Environment { val asString = "integration"  }
+  case object Development     extends Environment { val asString = "development"  }
+
+  val values: List[Environment] = List(Production, ExternalTest, Staging, QA, Integration, Development)
+
+  def parse(s: String): Option[Environment] =
+    values.find(_.asString == s)
```
cjwebb: Is it worth doing a `.toLowerCase` here?

app/uk/gov/hmrc/cataloguefrontend/service/ConfigService.scala
```diff
@@ -43,6 +43,7 @@ object ConfigService {
   case class ConfigSourceEntries(source: String, precedence: Int, entries: Map[KeyName, String] = Map())
   case class ConfigSourceValue(source: String, precedence: Int, value: String)
 
+  // TODO another variant to cater for - Environment + "local"...
```
cjwebb: What does this comment mean?

app/uk/gov/hmrc/cataloguefrontend/model/Environment.scala
```diff
@@ -0,0 +1,61 @@
...
+
+object Environment {
+  case object Production      extends Environment { val asString = "production"   }
+  case object ExternalTest    extends Environment { val asString = "externaltest" }
+  case object Staging         extends Environment { val asString = "staging"      }
+  case object QA              extends Environment { val asString = "qa"           }
+  case object Integration     extends Environment { val asString = "integration"  }
+  case object Development     extends Environment { val asString = "development"  }
+
+  val values: List[Environment] = List(Production, ExternalTest, Staging, QA, Integration, Development)
```
cjwebb: Does this List also double as an implicit ordering for lots of catalogue pages?
I wonder if we should explicitly define one.

app/uk/gov/hmrc/cataloguefrontend/model/Environment.scala
colin-lamed: TeamsAndRepositories also has spaces e.g. `"External Test"`. I think being restrictive means we actually know what the inputs from different boundaries look like, rather than going through a load of transformations.

app/uk/gov/hmrc/cataloguefrontend/service/ConfigService.scala
colin-lamed: I have further work to add Environment + Latest, and Environment + Latest + GithubMaster types. This is another case I don't want to loose.

app/uk/gov/hmrc/cataloguefrontend/model/Environment.scala
colin-lamed: Yes, it does. I'll add an Ordering to make this official/explicit.

PR #171, opened 2020-02-06
Bdog 609

colin-lamed:
Now that service-dependencies returns the group of a dependency, we can correct the link to the bobby rule by taking this into account.
Further, if there is more than one bobby rule for the given dependency, we pick one to link to by choosing the latest upper bound. This is consistent with `sbt-bobby`.

### DIFF
--- a/app/uk/gov/hmrc/cataloguefrontend/DependencyReportController.scala
+++ b/app/uk/gov/hmrc/cataloguefrontend/DependencyReportController.scala
@@ -107,8 +107,8 @@ class DependencyReportController @Inject()(
       case Some(VersionState.MinorVersionOutOfDate) => "amber"
       case Some(VersionState.MajorVersionOutOfDate) => "red"
       case Some(VersionState.Invalid)               => "N/A"
-      case Some(VersionState.BobbyRuleViolated)     => "verybad"
-      case Some(VersionState.BobbyRulePending)      => "pending"
+      case Some(_: VersionState.BobbyRuleViolated)  => "verybad"
+      case Some(_: VersionState.BobbyRulePending)   => "pending"
       case None                                     => "grey"
     }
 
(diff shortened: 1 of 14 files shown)


app/uk/gov/hmrc/cataloguefrontend/DependencyReportController.scala
```diff
@@ -107,8 +107,8 @@ class DependencyReportController @Inject()(
       case Some(VersionState.MinorVersionOutOfDate) => "amber"
       case Some(VersionState.MajorVersionOutOfDate) => "red"
       case Some(VersionState.Invalid)               => "N/A"
-      case Some(VersionState.BobbyRuleViolated)     => "verybad"
-      case Some(VersionState.BobbyRulePending)      => "pending"
+      case Some(_: VersionState.BobbyRuleViolated)  => "verybad"
+      case Some(_: VersionState.BobbyRulePending)   => "pending"
```
bennetimo: Minor but wondering if we should rename this to `BobbyRuleWarning` to match `sbt-bobby`? Also sounds a bit like it's waiting to be create or something

app/uk/gov/hmrc/cataloguefrontend/connector/model/Dependencies.scala
```diff
@@ -28,25 +28,64 @@ object VersionState {
   case object MinorVersionOutOfDate extends VersionState
   case object MajorVersionOutOfDate extends VersionState
   case object Invalid               extends VersionState
-  case object BobbyRuleViolated     extends VersionState
-  case object BobbyRulePending      extends VersionState
+  case class BobbyRuleViolated(violation: BobbyRuleViolation) extends VersionState
+  case class BobbyRulePending (violation: BobbyRuleViolation) extends VersionState
```
bennetimo: Should the `BobbyRuleViolation` be just `BobbyRule` maybe? Might be a bit clearer to read, that a violation or warning has an inner rule, instead of a warning that has an inner violation?

PR #172, opened 2020-02-10
BDOG-590 uses environment model for whats-running-where

christopherjturner:


app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/ReleasesConnector.scala
```diff
@@ -62,7 +60,19 @@ class ReleasesConnector @Inject()(http: HttpClient,
...
+    http
+      .GET[Seq[ServiceDeployment]](baseUrl, params)
+      .recover {
+        case NonFatal(ex) =>
+          Logger.error(s"An error occurred when connecting to $baseUrl: ${ex.getMessage}", ex)
+          Seq.empty
+      }
+  }
+
+  def heritageReleases(profile: Option[Profile])(implicit hc: HeaderCarrier): Future[Seq[ServiceDeployment]] = {
```
colin-lamed: Since this is the same as ecsReleases apart from the url, is it worth creating an internal one with the url as a param?

app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/WhatsRunningWhereController.scala
```diff
@@ -67,41 +69,55 @@ class WhatsRunningWhereController @Inject()(
...
+    }
+
+  def ecsAndHeritage: Action[AnyContent] =
+    Action.async { implicit request =>
+      for {
+        form <- Future.successful(WhatsRunningWhereFilter.form.bindFromRequest)
+        profile             = profileFrom(form)
+        selectedProfileType = form.fold(_ => None, _.profileType).getOrElse(ProfileType.Team)
+        (ecsReleases, profiles) <- (service.ecsReleases(profile), service.profiles).mapN { case (r, p) => (r, p) }
+        heritageReleases        <- service.releases(profile)
```
colin-lamed: move into the previous `mapN`?

app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/WhatsRunningWhereController.scala
```diff
@@ -67,41 +69,55 @@ class WhatsRunningWhereController @Inject()(
...
+        releases     = (ecsReleases ++ heritageReleases).sortBy(_.applicationName.asString)
+        environments = distinctEnvironments(releases)
+        profileNames = profiles.filter(_.profileType == selectedProfileType).map(_.profileName).sorted
+      } yield Ok(page(environments, releases, selectedProfileType, profileNames, form))
     }
 }
```
colin-lamed: Is it worth refactoring since `heritage`, `ecs` and `ecsAndHeritage` are so similar, except the additional `service.releases(profile)`?
E.g. heritage = ecsAndHeritage(ecs = false, heritage = true)
with
`if (ecs) service.ecsReleases else Future(Seq.empty)` etc.

app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/model.scala
```diff
@@ -20,16 +20,19 @@ import java.time.{LocalDateTime, ZoneOffset}
...
 import play.api.libs.json._
+import uk.gov.hmrc.cataloguefrontend.model.Environment
+import uk.gov.hmrc.cataloguefrontend.model.Environment._
 
-case class WhatsRunningWhere(
-                              applicationName: ServiceName,
-                              versions       : List[WhatsRunningWhereVersion])
+sealed trait DeploymentGeneration
+case object ECS extends DeploymentGeneration
+case object Heritage extends DeploymentGeneration
```
colin-lamed: add ECS and Heritage to `object DeploymentGeneration` namespace?

app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/model.scala
```diff
@@ -120,19 +135,15 @@ object TimeSeen {
...
-    // use `contains` so environments with prefixes/suffixes are sorted too
-    case x if x.contains("production")   => 0
-    case x if x.contains("externaltest") => 1
-    case x if x.contains("staging")      => 2
-    case x if x.contains("qa")           => 3
-    case x if x.contains("integration")  => 4
-    case x if x.contains("development")  => 5
-    case _                               => 6
+object WrWEnvironment {
+  private def precedence(environment: Environment): Int = environment match {
```
colin-lamed: Can we confirm we still need this? The default `Environment.ordering` should suffice - but it we really want reverse ordering, then = `val ordering = Environment.ordering.reverse`

cjwebb: I think there's an Ordering on Environment now...
WRW should just use the same one (and it will flip what's shown)

app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/WhatsRunningWhereController.scala
```diff
@@ -67,41 +69,55 @@ class WhatsRunningWhereController @Inject()(
...
-        environments         =  distinctEnvironments(releases)
-        profileNames         =  profiles.filter(_.profileType == selectedProfileType).map(_.profileName).sorted
-      } yield
-        Ok(page(environments, releases, selectedProfileType, profileNames, form))
+        environments = distinctEnvironments(releases)
+        profileNames = profiles.filter(_.profileType == selectedProfileType).map(_.profileName).sorted
+      } yield Ok(page(environments, releases, selectedProfileType, profileNames, form))
+    }
+
+  def ecsAndHeritage: Action[AnyContent] =
```
cjwebb: Remove?

app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/WhatsRunningWhereController.scala
christopherjturner: makes sense ecsAndHeritage isn't used (as other colin commented) seems like a good way of removing it but retaining functionality for the future

app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/ReleasesConnector.scala
christopherjturner: turns out this call wasn't used so i'll remove it

app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/WhatsRunningWhereController.scala
christopherjturner: removed

app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/model.scala
christopherjturner: I'll leave it with the default ordering until we hear otherwise :man_shrugging:

PR #185, opened 2020-06-05
BDOG-594 refactor whatrunningwhere endpoints to use Platform flag

christopherjturner:


app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/ReleasesConnector.scala
```diff
@@ -78,10 +78,24 @@ class ReleasesConnector @Inject()(
       }
   }
 
+  def ecsWhatsRunningWhere(profile: Option[Profile])(implicit hc: HeaderCarrier): Future[Seq[WhatsRunningWhere]] = {
```
colin-lamed, line 81: The function names are confusing with
- `releases: Seq[WhatsRunningWhere]`
- `ecsReleases: Seq[ServiceDeployment]`
- `ecsWhatsRunningWhere: Seq[WhatsRunningWhere]`
- `releasesForService: WhatsRunningWhere`
The name "WhatsRunningWhere" is great for a UI feature, but doesn't map to a domain well, especially with REST apis.
However we can address the naming once we have all the concurrent tickets merged in.

colin-lamed: Do we need both releases and ecsWhatsRunningWhere now that the backend API has been combined?
maybe just `releases(platform: Platform): Seq[WhatsRunningWhere]`?

PR #186, opened 2020-06-08
BDOG-391: Added called to get archived repositories, and filter out o…

JoPintoPaul:
…f team page

### DIFF
--- a/app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
+++ b/app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
@@ -99,7 +99,7 @@ class CatalogueController @Inject()(
     configuration.getOptional[String]("teams-and-repositories.link-name.jenkins").getOrElse("jenkins")
 
   private lazy val hideArchivedRepositoriesFromTeam: Boolean =
-    configuration.getOptional[Boolean]("team.hideArchivedRepositories").getOrElse(false)
+    configuration.get[Boolean]("team.hideArchivedRepositories")
 
   private val repoTypeToDetailsUrl = Map(
     RepoType.Service   -> routes.CatalogueController.service _,
@@ -229,9 +229,9 @@ class CatalogueController @Inject()(
             leakDetectionService.repositoriesWithLeaks,
             serviceDependencyConnector.dependenciesForTeam(teamName),
             serviceDependencyConnector.getCuratedSlugDependenciesForTeam(teamName, SlugInfoFlag.ForEnvironment(Environment.Production)),
-            if (hideArchivedRepositoriesFromTeam) teamsAndRepositoriesConnector.archivedRepositories else Future.successful(Nil)).mapN {
-            (teamMembers, teamDetails, reposWithLeaks, masterTeamDependencies, prodDependencies, archivedRepos) =>
-              val reposToHide: Seq[String] = archivedRepos.map(_.name)
+            if (hideArchivedRepositoriesFromTeam) teamsAndRepositoriesConnector.archivedRepositories.map(_.map(_.name)) else Future.successful(Nil))
+            .mapN {
+            (teamMembers, teamDetails, reposWithLeaks, masterTeamDependencies, prodDependencies, reposToHide) =>
               Ok(
                 teamInfoPage(
                   teamName               = teamInfo.name,
@@ -245,7 +245,7 @@ class CatalogueController @Inject()(
                   leaksFoundForTeam      = leakDetectionService.teamHasLeaks(teamInfo, reposWithLeaks),
                   hasLeaks               = leakDetectionService.hasLeaks(reposWithLeaks),
                   masterTeamDependencies = masterTeamDependencies.filterNot(repo => reposToHide.contains(repo.repositoryName)),
-                  prodDependencies       = prodDependencies.filterNot { case (service, _) => reposToHide.contains(service) }
+                  prodDependencies       = prodDependencies
                 )
               )
           }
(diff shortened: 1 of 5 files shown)


app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
```diff
@@ -98,6 +98,9 @@ class CatalogueController @Inject()(
   private lazy val jenkinsLinkName =
     configuration.getOptional[String]("teams-and-repositories.link-name.jenkins").getOrElse("jenkins")
 
+  private lazy val hideArchivedRepositoriesFromTeam: Boolean =
+    configuration.getOptional[Boolean]("team.hideArchivedRepositories").getOrElse(false)
```
colin-lamed, line 102: It's preferable to put all default values in config - i.e. it's not optional. That way you don't need to look through code to see what the behaviour would be without it, and config-explorer would also display it (although it's not wired up for platapps).

app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
```diff
@@ -225,20 +228,24 @@ class CatalogueController @Inject()(
             userManagementConnector.getTeamDetails(teamName),
             leakDetectionService.repositoriesWithLeaks,
             serviceDependencyConnector.dependenciesForTeam(teamName),
-            serviceDependencyConnector.getCuratedSlugDependenciesForTeam(teamName, SlugInfoFlag.ForEnvironment(Environment.Production))).mapN {
-            (teamMembers, teamDetails, reposWithLeaks, masterTeamDependencies, prodDependencies) =>
+            serviceDependencyConnector.getCuratedSlugDependenciesForTeam(teamName, SlugInfoFlag.ForEnvironment(Environment.Production)),
+            if (hideArchivedRepositoriesFromTeam) teamsAndRepositoriesConnector.archivedRepositories else Future.successful(Nil)).mapN {
+            (teamMembers, teamDetails, reposWithLeaks, masterTeamDependencies, prodDependencies, archivedRepos) =>
+              val reposToHide: Seq[String] = archivedRepos.map(_.name)
```
colin-lamed, line 234: could move the `.map(_.name)` above so the data is ready for `.mapN` i.e `.mapN {(.., reposToHide) => }`

app/uk/gov/hmrc/cataloguefrontend/connector/TeamsAndRepositoriesConnector.scala
```diff
@@ -212,33 +209,31 @@ class TeamsAndRepositoriesConnector @Inject()(
...
 
   def digitalServiceInfo(digitalServiceName: String)(implicit hc: HeaderCarrier): Future[Option[DigitalService]] = {
     val url = teamsAndServicesBaseUrl + s"/api/digital-services/${encodePathParam(digitalServiceName)}"
     http.GET[Option[DigitalService]](url)
   }
 
   def allRepositories(implicit hc: HeaderCarrier): Future[Seq[RepositoryDisplayDetails]] =
+    http.GET[Seq[RepositoryDisplayDetails]](teamsAndServicesBaseUrl + s"/api/repositories")
+
+  def archivedRepositories(implicit hc: HeaderCarrier): Future[Seq[RepositoryDisplayDetails]] =
```
colin-lamed, line 232: could create a func `def repositories(archived: Option[Boolean])` then can delegate `allRepositories` and `archivedRepositories` to it (or even just call directly passing the relevant `archived` param)

app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
colin-lamed: just seen that there is already a default in `application.conf`

app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
```diff
@@ -225,20 +228,24 @@ class CatalogueController @Inject()(
...
                   activityDates          = TeamActivityDates(teamInfo.firstActiveDate, teamInfo.lastActiveDate, teamInfo.firstServiceCreationDate),
                   errorOrTeamMembers     = convertToDisplayableTeamMembers(teamInfo.name, teamMembers),
                   errorOrTeamDetails     = teamDetails,
                   umpMyTeamsUrl          = umpMyTeamsPageUrl(teamInfo.name),
                   leaksFoundForTeam      = leakDetectionService.teamHasLeaks(teamInfo, reposWithLeaks),
                   hasLeaks               = leakDetectionService.hasLeaks(reposWithLeaks),
-                  masterTeamDependencies = masterTeamDependencies,
-                  prodDependencies       = prodDependencies
+                  masterTeamDependencies = masterTeamDependencies.filterNot(dependencies => reposToHide.contains(dependencies.repositoryName)),
+                  prodDependencies       = prodDependencies.filterNot { case (repo, _) => reposToHide.contains(repo) }
```
colin-lamed, line 248: maybe don't need to filter `prodDependencies` - if it's in production we would want to know about it

app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
JoPintoPaul: @colin-lamed Addressed this.

app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
JoPintoPaul: Done.

app/uk/gov/hmrc/cataloguefrontend/connector/TeamsAndRepositoriesConnector.scala
JoPintoPaul: Done. I prefer to keep it as two functions externally as I think it is more self-documenting that way.

app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
JoPintoPaul: Good point. Done.

PR #211, opened 2020-11-25
BDOG-1134

Tyrpix:


app/views/DeploymentHistoryPage.scala.html
```diff
@@ -51,7 +51,7 @@ <h1>Deployments</h1>
                 <input type="hidden" id="selected-environment" value="@selectedEnv.asString">
                 <div class="col-xs-3">
                     <label for="search" title="Search across all fields" data-toggle="tooltip">Search</label>
-                    <input id="search" class="search form-control" type="text" name="search" value='@form("search").value' autofocus>
+                    <input id="search" class="search form-control" type="text" name="search" value='@form("search").value' autofocus autocomplete="off">
```
cjwebb, line 54: why autocomplete=off?

app/views/DeploymentHistoryPage.scala.html
```diff
@@ -130,6 +134,12 @@ <h1>Deployments</h1>
 
 var searchBox = document.getElementById("search");
 
+serviceList.on('searchComplete', function(){
+  var numberOfDeployments = serviceList.update().matchingItems.length;
```
cjwebb, line 138: What does the `update()` here do?
What happens if you just have `serviceList.matchingItems.length`?

The update may indeed be required; I'd just like to understand the behavior.

app/views/DeploymentHistoryPage.scala.html
Tyrpix: It stops the drop down of the search box covering the number of deployments

christopherjturner (approved):

PR #230, opened 2021-03-08
Bdog 1299

JasperWagen:


app/uk/gov/hmrc/cataloguefrontend/healthindicators/HealthIndicatorsConnector.scala
```diff
@@ -32,14 +32,21 @@ class HealthIndicatorsConnector @Inject()(
   import HttpReads.Implicits._
   private val logger = Logger(getClass)
 
+  implicit val rrR: Reads[RepositoryRating] = RepositoryRating.reads
```
nigelhp, line 35: Scope - this should not be public.
As a top-level member it deserves a better name (even though you only use it implicitly).

app/uk/gov/hmrc/cataloguefrontend/healthindicators/HealthIndicatorsConnector.scala
```diff
@@ -32,14 +32,21 @@ class HealthIndicatorsConnector @Inject()(
...
+
   private val healthIndicatorsBaseUrl: String = servicesConfig.baseUrl("health-indicators")
 
   def getHealthIndicators(repoName: String)(implicit hc: HeaderCarrier): Future[Option[RepositoryRating]] = {
-    implicit val rrR: Reads[RepositoryRating] = RepositoryRating.reads
     val url                                   = s"$healthIndicatorsBaseUrl/health-indicators/repositories/$repoName"
```
nigelhp, line 40: Formatting is all out of whack now that the other variable has been removed.
This is one of the issues with the "align fields" style  ...

app/uk/gov/hmrc/cataloguefrontend/healthindicators/HealthIndicatorsConnector.scala
```diff
@@ -87,12 +94,13 @@ object Rating {
   }
 }
 
-case class RepositoryRating(repositoryName: String, repositoryScore: Int, ratings: Seq[Rating])
+case class RepositoryRating(repositoryName: String, repositoryType: String, repositoryScore: Int, ratings: Seq[Rating])
```
nigelhp, line 97: Try and avoid `String` where an enumerated type would be preferable.
A `RepoType` classification is used below.  Can that be used here?

nigelhp (approved):

nigelhp (approved):

PR #250, opened 2021-06-22
BDOG-1348: Refactors Naming

Tyrpix:


app/uk/gov/hmrc/cataloguefrontend/healthindicators/HealthIndicatorsConnector.scala
```diff
@@ -30,88 +30,88 @@ class HealthIndicatorsConnector @Inject()(
...
     http
-      .GET[Option[RepositoryRating]](url)
+      .GET[Option[Indicator]](url)
   }
 
-  def getAllRepositoryRatings(repoType: RepoType)(implicit hc: HeaderCarrier): Future[Seq[RepositoryRating]] = {
+  def getAllIndicators(repoType: RepoType)(implicit hc: HeaderCarrier): Future[Seq[Indicator]] = {
     // AllTypes is represented on the backend by sending no RepoType parameter
     val repoTypeQueryP = if(repoType == RepoType.AllTypes) "" else "&repoType=" + repoType.asString
     val url = s"$healthIndicatorsBaseUrl/health-indicators/repositories/?sort=desc" ++ repoTypeQueryP
```
anshulbajpai, line 46: It is unrelated to your change but I think we should not create a URL like this by string interpolation. The `GET` method accepts a list of query params. We can pass the query params as the argument to that parameter.

anshulbajpai: Can you make this change please?

app/uk/gov/hmrc/cataloguefrontend/healthindicators/HealthIndicatorsConnector.scala
```diff
@@ -30,88 +30,88 @@ class HealthIndicatorsConnector @Inject()(
...
+        case "leak-detection"  => JsSuccess(LeakDetection)
+        case "bobby-rule"      => JsSuccess(BobbyRule)
+        case "build-stability" => JsSuccess(BuildStability)
+        case "alert-config"    => JsSuccess(AlertConfig)
+        case "open-pr"         => JsSuccess(OpenPR)
+        case s                => JsError(s"Invalid MetricType: $s")
       }
 
-    override def writes(o: RatingType): JsValue =
+    override def writes(o: MetricType): JsValue =
```
anshulbajpai, line 66: Why do we have a writes? Are we not only reading from the API?

app/uk/gov/hmrc/cataloguefrontend/healthindicators/HealthIndicatorsService.scala
```diff
@@ -31,24 +32,24 @@ class HealthIndicatorsService @Inject()(
...
+      indicators  <- eventualIndicators
     } yield
-      allRepoRatings.map { rr =>
-        RepoRatingsWithTeams(
-          rr.repositoryName,
-          owningTeams = repoToTeams.getOrElse(rr.repositoryName, Seq.empty),
-          rr.repositoryType,
-          rr.repositoryScore,
-          rr.ratings
+      indicators.map { rr =>
```
anshulbajpai, line 43: rr?

anshulbajpai (approved):

PR #286, opened 2021-09-13
integrates with platform metrics v2

shah-ee:


christopherjturner (approved):

app/uk/gov/hmrc/cataloguefrontend/metrics/model/ServiceMetricsEntry.scala
```diff
@@ -16,9 +16,14 @@
...
 package uk.gov.hmrc.cataloguefrontend.metrics.model
 
-final case class MetricsEntry (dependency: DependencyName, happyCount: Int, unHappyCount: Int){
+case class ServiceMetricsEntry(
+                           name: String,
+                           count: Int,
+                           happyCount: Int,
+                           unHappyCount: Int
+                         ){
   def asPercentage: List[Int] = {
```
colin-lamed, line 25: It's not intuitive what the returned List is. Would it be better to add derived fields`def happyPercent` `def unhappyPercent`?
Do we need count as well as happyCount/unhappyCount in the API - it can all be derived and adds room for being out-of-sync?

app/uk/gov/hmrc/cataloguefrontend/metrics/model/MetricsResponse.scala
```diff
@@ -16,10 +16,10 @@
...
 
-import play.api.libs.json.{Json, Reads}
+import play.api.libs.json.Reads
+
+final case class MetricsResponse(metrics: Seq[ServiceProgressMetrics])
 
-case class MetricsResponse(metrics: Seq[ServiceProgressMetrics])
 object MetricsResponse {
-  implicit val reads: Reads[MetricsResponse] = Json.reads[MetricsResponse]
+  implicit val reads: Reads[MetricsResponse] = implicitly[Reads[Seq[ServiceProgressMetrics]]].map(MetricsResponse.apply)
```
colin-lamed, line 24: Should this be implicit on the companion object, since it only applies to the `MetricsConnector`? Also couples the model field name "metrics" to the external API.

app/uk/gov/hmrc/cataloguefrontend/metrics/connector/MetricsConnector.scala
```diff
@@ -57,11 +49,9 @@ object MetricsConnector{
 
     import cats.syntax.option._
 
-    override def query(maybeGroup: Option[GroupName], maybeName: Option[DependencyName], maybeRepository: Option[RepositoryName]): Future[MetricsResponse] = {
+    override def query(maybeTeam: Option[TeamName]): Future[MetricsResponse] = {
       val url = url"$platformProgressMetricsBaseURL/platform-progress-metrics/metrics" + List(
-        maybeGroup.map(g => s"group=${g.value}").orEmpty,
-        maybeName.map(n => s"name=${n.value}").orEmpty,
-        maybeRepository.map(r => s"repository=${r.value}").orEmpty
+        maybeTeam.map(r => s"team=${r.asString}").orEmpty
```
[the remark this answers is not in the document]
shah-ee: > using the url interpolator
Ah cool! thanks didn't know if play had them. will do.

> team should probably be included in the results
> should repoType also be included in the results?

agreed. thought I'd do them in stages and have the full metrics available first. will do these in the next PR.

app/uk/gov/hmrc/cataloguefrontend/metrics/model/ServiceMetricsEntry.scala
shah-ee: > It's not intuitive what the returned List is. Would it be better to add derived fieldsdef happyPercent def unhappyPercent?

agreed. I initially had them as tuples but then I struggled with accessing them in the html template. can have them as separate methods and construct the list of pie values (happy%, unhappy%) explicitly. 

> Do we need count as well as happyCount/unhappyCount in the API? platform metrics does expose `count` which helps sanity check the results like you say. wasn't sure what the benefit of parsing them in catalogue-frontend would be tho. happy to add them in if you feel they'd be useful.

app/uk/gov/hmrc/cataloguefrontend/metrics/model/MetricsResponse.scala
shah-ee: > Should this be implicit on the companion object, since it only applies to the MetricsConnector?

we had this discussion. while I appreciate your point about coherent vs incoherent type classes, we have already limited this type to the `metrics` package. so not sure if we should worry about models being reused. 

- as for the naming `MetricsResponse`, in the metrics service, I split the models into: 
- internal API
- external API
- domain model

didn't bother with all that here. but to clarify it's actually the response we get from the metrics service, as opposed to the response we expose to the catalogue-frontend UI. it's just a wrapper type for the array. (I seem to have forgotten to add it as an `AnyVal`)  what do you suggest we call it?

PR #288, opened 2021-09-15
filtering metrics by teams in metrics overview

shah-ee:


app/uk/gov/hmrc/cataloguefrontend/metrics/connector/MetricsConnector.scala
```diff
@@ -41,23 +39,31 @@ object MetricsConnector{
...
-        .recoverWith {
+        .onError {
           case UpstreamErrorResponse.Upstream5xxResponse(x) =>
-            logger.error(s"An error occurred when connecting to platform progress metrics service. baseUrl: $platformProgressMetricsBaseURL", x)
-            Future.successful(MetricsResponse(Seq.empty))
+            Future.successful(
+              logger.error(s"An error occurred when connecting to platform progress metrics service. baseUrl: $platformProgressMetricsBaseURL", x)
+            )
+        }
+        .recover { case _ =>
```
colin-lamed, line 65: Non 5xx errors would not be logged, and just suppressed? I think it's better to log and recover together (as previously).
It's also arguable that it's better to fail the page rather than return misleading results.

shah-ee: > Non 5xx errors would not be logged, and just suppressed? 

Good point! it was a copy paste. will fix.

> think it's better to log and recover together.

I believe keeping the recovery logic pure and separating effects such as logging through the likes of `onError` and `flatTap` is considered to be best practice from a FP point of view. 

> It's also arguable that it's better to fail the page rather than return misleading results.

sure. do we want to distinguish between 500 and non-500 ones? :thinking: I have not done a whole lot of FE, but my understanding is that folks tend to only expose 500s and non-500 ones if you can actually do something about it.

colin-lamed: We already catch all unhandled exceptions, log, and present an error page.
If you're not recovering, we probably don't need the log - we'll just get double logging (with stack trace) - if you want to augment the message, something like `recoverWith { case e: Exception => Future.failed(new RuntimeException("...", e)) }`.

colin-lamed (approved):

PR #290, opened 2021-09-27
BDOG-1477 adds additional search functionality and counter to default…

TimothyFothergill:
… branch names page.

### DIFF
--- a/app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
+++ b/app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
@@ -498,7 +498,7 @@ class CatalogueController @Inject() (
       }
     }
 
-  def allDefaultBranches(singleOwnership: Boolean, archived: Boolean): Action[AnyContent] = {
+  def allDefaultBranches(singleOwnership: Boolean, includeArchived: Boolean): Action[AnyContent] = {
     Action.async { implicit request =>
       teamsAndRepositoriesConnector.allDefaultBranches.map { repositories =>
         DefaultBranchesFilter.form
@@ -508,15 +508,15 @@ class CatalogueController @Inject() (
               repositories      = Seq(),
               teams             = Seq(""),
               singleOwnership   = false,
-              archived          = false,
+              archived   = false,
               formWithErrors)),
             query =>
               Ok(
                 defaultBranchListPage(
-                  repositories = defaultBranchesService.filterRepositories(repositories, query.name, query.defaultBranch, query.teamNames, singleOwnership, archived),
+                  repositories = defaultBranchesService.filterRepositories(repositories, query.name, query.defaultBranch, query.teamNames, singleOwnership, includeArchived),
                   teams = defaultBranchesService.allTeams(repositories),
                   singleOwnership = singleOwnership,
-                  archived = archived,
+                  archived = includeArchived,
                   DefaultBranchesFilter.form.bindFromRequest()
                 )
               )
--- a/app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
+++ b/app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
@@ -28,11 +28,11 @@ import uk.gov.hmrc.cataloguefrontend.DisplayableTeamMember._
 import uk.gov.hmrc.cataloguefrontend.actions.{UmpAuthActionBuilder, VerifySignInStatus}
 import uk.gov.hmrc.cataloguefrontend.connector._
 import uk.gov.hmrc.cataloguefrontend.connector.UserManagementConnector.UMPError
-import uk.gov.hmrc.cataloguefrontend.connector.model.{Dependency, DependencyScope, Version, TeamName}
+import uk.gov.hmrc.cataloguefrontend.connector.model.{Dependency, DependencyScope, TeamName, Version}
 import uk.gov.hmrc.cataloguefrontend.events._
 import uk.gov.hmrc.cataloguefrontend.model.{Environment, SlugInfoFlag}
 import uk.gov.hmrc.cataloguefrontend.service.ConfigService.ArtifactNameResult.{ArtifactNameError, ArtifactNameFound, ArtifactNameNotFound}
-import uk.gov.hmrc.cataloguefrontend.service.{ConfigService, LeakDetectionService, RouteRulesService}
+import uk.gov.hmrc.cataloguefrontend.service.{ConfigService, DefaultBranchesService, LeakDetectionService, RouteRulesService}
 import uk.gov.hmrc.cataloguefrontend.shuttering.{ShutterService, ShutterState, ShutterType}
 import uk.gov.hmrc.cataloguefrontend.util.MarkdownLoader
 import uk.gov.hmrc.cataloguefrontend.whatsrunningwhere.WhatsRunningWhereService
@@ -66,6 +66,7 @@ class CatalogueController @Inject() (
   eventService                 : EventService,
   readModelService             : ReadModelService,
   shutterService               : ShutterService,
+  defaultBranchesService       : DefaultBranchesService,
   verifySignInStatus           : VerifySignInStatus,
   umpAuthActionBuilder         : UmpAuthActionBuilder,
   userManagementPortalConfig   : UserManagementPortalConfig,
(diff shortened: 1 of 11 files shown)


app/uk/gov/hmrc/cataloguefrontend/service/DefaultBranchesService.scala
```diff
@@ -0,0 +1,71 @@
...
+      singleOwnership:  Option[Boolean],
+      archived:         Option[Boolean]): Int = {
+
+    var results: Seq[RepositoryDisplayDetails] = repos
+
+    if(name.isDefined)              { results = results.filter(_.name contains name.get) }
+    if(defaultBranch.isDefined)     { results = results.filter(_.defaultBranch contains defaultBranch.get) }
+    if(teamNames.isDefined)         { results = results.filter(_.teamNames contains teamNames.get) }
+    if(singleOwnership.isDefined)   { results = results.filter(_.teamNames.length == 1) }
+    if(archived.isEmpty)            { results = results.filter(_.isArchived == false) }
```
christopherjturner, line 48: are these filters exclusive? if so then `else if` might be better than multiple ifs. 
If they're not exclusive is `results` just being overwritten each time?

TimothyFothergill: They're exclusive - This is because each search field on the page triggers a separate if statement. `results` is being overwritten each time still.

app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
```diff
@@ -497,24 +498,28 @@ class CatalogueController @Inject() (
...
-  def allDefaultBranches: Action[AnyContent] =
+  def allDefaultBranches(singleOwnership: Option[Boolean], archived: Option[Boolean]): Action[AnyContent] = {
     Action.async { implicit request =>
-
       teamsAndRepositoriesConnector.allDefaultBranches.map { repositories =>
         DefaultBranchesFilter.form
           .bindFromRequest()
           .fold(
-            formWithErrors => Ok(defaultBranchListPage(repositories = Seq.empty, formWithErrors)),
+            formWithErrors => Ok(defaultBranchListPage(repositories = Seq(), calculate = 0, teams = Seq(""), Option(false), Option(false), formWithErrors)),
```
colin-lamed, line 507: use named params for values that need context (`Option(false)`)

christopherjturner (dismissed):

app/uk/gov/hmrc/cataloguefrontend/service/DefaultBranchesService.scala
colin-lamed: what if `archived = Some(true)` - shouldn't it be `filter(_.isArchived == archived.get)`?
By creating a pattern, this can be simplified too.

app/uk/gov/hmrc/cataloguefrontend/service/DefaultBranchesService.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  def updatedDefaultBranchCount(
+      repos:            Seq[RepositoryDisplayDetails],
+      name:             Option[String],
+      defaultBranch:    Option[String],
+      teamNames:        Option[String],
+      singleOwnership:  Option[Boolean],
+      archived:         Option[Boolean]): Int = {
+
+    var results: Seq[RepositoryDisplayDetails] = repos
```
colin-lamed, line 42: can we rewrite this to avoid the `var`? Also avoid using `if (opt.isDefined) opt.get` pattern (which can break with refactoring).
Just one idea:
```scala
implicit class Syntax(repos: Seq[RepositoryDisplayDetails]) {
  def filterBy[A](optA: Option[A], f: RepositoryDisplayDetails => A => Boolean): Seq[RepositoryDisplayDetails] =
    optA.fold(repos)(a => repos.filter(repo => f(repo, a))
}
val results =
  repos
    .filterBy(name, _.name contains _)
    .filterBy(singleOwnership)(_ => _.teamNames.length == 1)
```

app/uk/gov/hmrc/cataloguefrontend/service/DefaultBranchesService.scala
```diff
@@ -0,0 +1,71 @@
...
+
+@Singleton
+class DefaultBranchesService @Inject()()(implicit val ec: ExecutionContext){
+
+  def allTeams(repos: Seq[RepositoryDisplayDetails]): Seq[String] = {
+    val repositories = repos.map(t => t.teamNames)
+    repositories.flatten.distinct.sorted
+  }
+
+  def updatedDefaultBranchCount(
```
colin-lamed, line 34: Can you not just use `filterRepositories(..).count(_.defaultCount != "master"` to avoid duplicating all the logic?

app/views/DefaultBranchListPage.scala.html
```diff
@@ -17,18 +17,24 @@
...
+@import uk.gov.hmrc.cataloguefrontend.service.DefaultBranchesService
 @import uk.gov.hmrc.cataloguefrontend.ViewMessages
 @import helper._
+@import scala.concurrent.Future
 
 @this(viewMessages: ViewMessages)
 
 @(repositories: Seq[RepositoryDisplayDetails],
-form: Form[_]
+    calculate: Int,
```
colin-lamed, line 28: Should this be named something more explicit, like `updatedDefaultBranchCount`?
You are already passing `repositories` in, I don't think we need `calculate` as well - It's simpler to understand `@{repositories.filter(_.defaultBranch != "master).count} / @{repositories.count}`

app/views/DefaultBranchListPage.scala.html
```diff
@@ -44,32 +50,51 @@ <h1>Default Branch</h1>
...
-                <label for="search">Name:</label>
-            </div>
             <div class="col col-xs-3">
-                <input class="search form-control" id="search" type="text" name="name" value='@form("name").value' autofocus>
-            </div>
-            <div class="col-xs-1 padding-reset-right">
-                <label for="teamNames">Team:</label>
+                <label for="search">Name:</label>
+                <div>
+                    <input class="form-control" id="search" type="text" name="name" value='@form("name").value'>
```
colin-lamed, line 56: there are form functions for displaying input fields (for text checkbox etc) which should make it simpler to sync to the form value.

app/views/DefaultBranchListPage.scala.html
```diff
@@ -79,28 +104,29 @@ <h1>Default Branch</h1>
             </td>
             <td class="teamNames monospaced" id="row@{i}_teamNames">
                 @if(repo.teamNames.length > 0 && repo.teamNames.length <= 6) { @for(name <- repo.teamNames) {
-                            @if(name == repo.teamNames.last) { @name } else { @name, }
+                    @if(name == repo.teamNames.last) { @name } else { @name, }
                         }
                     } else if (repo.teamNames.length > 6)  {
-                        <i>Various Teams</i>
+                        <i title="@for(name <- repo.teamNames) { @if(name == repo.teamNames.last) { @name } else { @name,}} ">Various Teams</i>
```
colin-lamed, line 110: you can just use `mkString`: `repos.map(_.teamNames).mkString(",")`
Same above.

app/uk/gov/hmrc/cataloguefrontend/service/DefaultBranchesService.scala
colin-lamed: Or better (also with singleOwnership and archived as Boolean rather than a three state Option[Boolean])
```scala
repos
  .filter(repo => optName.fold(true)(repo._name contains _)
  .filter(repo => !singleOwnership || repo.timeNames.length == 1)
  .filter(repo => !onlyNonArchived || !repo.isArchived)
```

app/views/DefaultBranchListPage.scala.html
```diff
@@ -107,7 +106,7 @@ <h1>Default Branch Naming Progress</h1>
                     @if(name == repo.teamNames.last) { @name } else { @name, }
```
colin-lamed, line 106: mkString can be used here too

app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
```diff
@@ -498,18 +498,22 @@ class CatalogueController @Inject() (
       }
     }
 
-  def allDefaultBranches(singleOwnership: Option[Boolean], archived: Option[Boolean]): Action[AnyContent] = {
+  def allDefaultBranches(singleOwnership: Boolean, archived: Boolean): Action[AnyContent] = {
```
colin-lamed, line 501: can we rename `archived` to `includeArchived`? otherwise you may expect `archived=true` would only return archived repositories.

app/views/DefaultBranchListPage.scala.html
```diff
@@ -85,8 +84,8 @@ <h1>Default Branch Naming Progress</h1>
     </form>
     <table class="table table-striped">
         <div>
-        <td colspan="4"><progress class="progress" style="width: 100%;" id="nomenclature" max="@repositories.length" value = @calculate> @calculate </progress> </td>
-            Updated default branch names: @calculate/@repositories.length
+        <td colspan="4"><progress class="progress" style="width: 100%;" id="nomenclature" max="@repositories.length" value='@repositories.filter(_.defaultBranch != "master").length'> @repositories.filter(_.defaultBranch != "master").length </progress> </td>
```
colin-lamed, line 87: I see you need the number of non-master repos twice - you could use `@defining`
```scala
@defining(repositories.filter(_.defaultBranch != "master").length) { nonMasterCount =>
  @nonMasterCount/@repositories.length
}
```

colin-lamed (approved):

colin-lamed (dismissed):

colin-lamed (approved):

PR #314, opened 2021-12-13
Provide service cost estimates

tomwadeson:
Estimated yearly cost (total ~slots~ instances across all environments * ~£650.00~ $120) is now displayed on the `ServiceInfoPage`, e.g. on `/repositories/repository-name` where `repository-name` is the name of a service.

Here's an example view:

Screenshot 2021-12-13 at 16 10 25

--

I've still not been able to spin this up locally with an export of the Mongo DB in order to kick the tyres properly, so I'll hold off on merging this until I can do that (waiting for an automated job to fire tomorrow morning to finalise my access.)

### DIFF
--- a/app/uk/gov/hmrc/cataloguefrontend/connector/ConfigConnector.scala
+++ b/app/uk/gov/hmrc/cataloguefrontend/connector/ConfigConnector.scala
@@ -55,8 +55,6 @@ class ConfigConnector @Inject() (
   )(implicit hc: HeaderCarrier): Future[Option[DeploymentConfig]] = {
     implicit val dcr = DeploymentConfig.reads
     http
-      .GET[DeploymentConfig](s"$serviceConfigsBaseUrl/deployment-config/${environment.asString}/$service")
-      .map(Some(_))
-      .recover { case _ => None }
+      .GET[Option[DeploymentConfig]](s"$serviceConfigsBaseUrl/deployment-config/${environment.asString}/$service")
   }
 }
--- a/app/uk/gov/hmrc/cataloguefrontend/connector/ConfigConnector.scala
+++ b/app/uk/gov/hmrc/cataloguefrontend/connector/ConfigConnector.scala
@@ -21,7 +21,7 @@ import uk.gov.hmrc.cataloguefrontend.connector.model.BobbyRuleSet
 import uk.gov.hmrc.cataloguefrontend.model.Environment
 import uk.gov.hmrc.cataloguefrontend.service.ConfigService._
 import uk.gov.hmrc.cataloguefrontend.service.CostEstimationService.DeploymentConfig
-import uk.gov.hmrc.http.{HeaderCarrier, HttpClient, HttpReads}
+import uk.gov.hmrc.http.{HeaderCarrier, HttpClient, HttpReads, StringContextOps}
 import uk.gov.hmrc.play.bootstrap.config.ServicesConfig
 
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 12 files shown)


app/uk/gov/hmrc/cataloguefrontend/connector/ConfigConnector.scala
```diff
@@ -46,4 +48,15 @@ class ConfigConnector @Inject() (
...
+
+  def deploymentConfig(
+    service: String,
+    environment: Environment
+  )(implicit hc: HeaderCarrier): Future[Option[DeploymentConfig]] = {
+    implicit val dcr = DeploymentConfig.reads
+    http
+      .GET[DeploymentConfig](s"$serviceConfigsBaseUrl/deployment-config/${environment.asString}/$service")
+      .map(Some(_))
+      .recover { case _ => None }
```
tomwadeson, line 60: This is quite blunt. Would we prefer to log the error here before returning `None`?

app/uk/gov/hmrc/cataloguefrontend/service/CostEstimationService.scala
```diff
@@ -0,0 +1,80 @@
...
+
+@Singleton
+final class CostEstimationService @Inject() (configConnector: ConfigConnector) {
+
+  def estimateServiceCost(
+    service: String,
+    environments: Seq[Environment]
+  )(implicit ec: ExecutionContext, hc: HeaderCarrier): Future[CostEstimation] =
+    Future
+      .traverse(environments)(environment =>
```
tomwadeson, line 37: Here we're making a call to `service-configs` for each environment that the service runs in. Happy to revisit this and expose a new API in `service-configs` to fetch all configs for all environments if we'd prefer to do that though.

app/uk/gov/hmrc/cataloguefrontend/connector/ConfigConnector.scala
colin-lamed: I think it should at least log the error.
This function could be used in other contexts, so I think propagating the error is best - the api is internal, and we don't expect it to be down. `None` usually means there is no config, so it could be confusing. - note, if we're expecting to handle a 404 as `None` it should be `.GET[Option[DeploymentConfig]]` - the implicit reads will handle the 404.

app/uk/gov/hmrc/cataloguefrontend/service/CostEstimationService.scala
```diff
@@ -0,0 +1,80 @@
...
+        configConnector
+          .deploymentConfig(service, environment)
+          .map(deploymentConfig => (environment, deploymentConfig.getOrElse(DeploymentConfig(0, 0))))
```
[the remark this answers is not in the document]
colin-lamed: Seems fine to me. maybe add named params to explain what we're setting to `0` - or even define a `DeploymentConfig.empty` if appropriate

app/uk/gov/hmrc/cataloguefrontend/service/CostEstimationService.scala
```diff
@@ -0,0 +1,80 @@
...
+  object DeploymentConfig {
+    val reads: Reads[DeploymentConfig] = Json.reads[DeploymentConfig]
+  }
+
+  final case class CostEstimation(totalSlots: Int) {
+
+    val yearlyCost: Double =
+      totalSlots * 650.0
+
+    val yearlyCostFormatted: String = {
```
colin-lamed, line 63: should derived fields be lazy? this one with a formatter could be expensive?
Wonder if the formatting be done in the view, since is a view concern?

app/uk/gov/hmrc/cataloguefrontend/connector/ConfigConnector.scala
tomwadeson: Ah, yes, `.GET[Option[DeploymentConfig]]` is exactly what I want here, thanks

app/uk/gov/hmrc/cataloguefrontend/service/CostEstimationService.scala
tomwadeson: I'll make them lazy.

I'm a little torn on moving it to the view. I'm with you that it feels like a view concern, but I do like being able to test it without asserting against the HTML. Is that persuasive for you at all? If not, happy to move it there.

colin-lamed: you could make a view utility? It would help ensure numbers are consistently displayed too

colin-lamed (dismissed):

colin-lamed (approved):

PR #326, opened 2022-01-10
Cost estimates include per-environment breakdowns

tomwadeson:
Looks a little something like this:
Screenshot 2022-01-10 at 16 04 18

### DIFF
--- a/app/uk/gov/hmrc/cataloguefrontend/service/CostEstimationService.scala
+++ b/app/uk/gov/hmrc/cataloguefrontend/service/CostEstimationService.scala
@@ -64,8 +64,7 @@ object CostEstimationService {
   final case class ServiceCostEstimate(forEnvironments: List[ServiceCostEstimate.ForEnvironment]) {
 
     lazy val summary: ServiceCostEstimate.Summary =
-      forEnvironments
-        .foldLeft(Summary.zero)((summary, forEnv) => summary + Summary(forEnv.slots, forEnv.yearlyCostGbp))
+      forEnvironments.map(_.summary).fold(Summary.zero)(_ + _)
   }
 
   object ServiceCostEstimate {
@@ -88,7 +87,11 @@ object CostEstimationService {
       environment: Environment,
       slots: Int,
       yearlyCostGbp: Double
-    )
+    ) {
+
+      def summary: Summary =
+        Summary(slots, yearlyCostGbp)
+    }
 
     final case class Summary(totalSlots: Int, totalYearlyCostGbp: Double) {
       def +(other: Summary): Summary =
--- a/app/uk/gov/hmrc/cataloguefrontend/service/CostEstimationService.scala
+++ b/app/uk/gov/hmrc/cataloguefrontend/service/CostEstimationService.scala
@@ -19,7 +19,8 @@ package uk.gov.hmrc.cataloguefrontend.service
 import play.api.libs.json.{Json, Reads}
 import uk.gov.hmrc.cataloguefrontend.connector.ConfigConnector
 import uk.gov.hmrc.cataloguefrontend.model.Environment
-import uk.gov.hmrc.cataloguefrontend.service.CostEstimationService.{CostEstimation, DeploymentConfig}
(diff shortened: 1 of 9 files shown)


app/views/partials/costEstimation.scala.html
```diff
@@ -14,21 +14,43 @@
...
+<script type="text/javascript" src="https://www.gstatic.com/charts/loader.js"></script>
+<script type="text/javascript">
+      google.charts.load("current", {packages:["corechart"]});
+      google.charts.setOnLoadCallback(drawChart);
+
+      function drawChart() {
+        var data = google.visualization.arrayToDataTable([
+          ['Environment',   'Estimated Cost'],
+          @for(e <- serviceCostEstimate.forEnvironments) {
+            ['@e.environment.asString', { v: @e.yearlyCostGbp, f: "@formatGbp(e.yearlyCostGbp) (slots = @e.slots)" } ],
```
tomwadeson, line 47: Is this super gross? Should I create a view model and associated JSON format to do this properly?

app/uk/gov/hmrc/cataloguefrontend/service/CostEstimationService.scala
```diff
@@ -58,21 +61,65 @@ object CostEstimationService {
       Json.reads[DeploymentConfig]
   }
 
-  final case class CostEstimation(totalSlots: Int) {
+  final case class ServiceCostEstimate(forEnvironments: List[ServiceCostEstimate.ForEnvironment]) {
```
colin-lamed, line 64: Would `Map[Environment,Summary]` be more practical? (we can be confident we only one for each etc.)

app/uk/gov/hmrc/cataloguefrontend/service/CostEstimationService.scala
```diff
@@ -58,21 +61,65 @@ object CostEstimationService {
...
 
-    lazy val yearlyCostGbp: Double =
-      totalSlots * 650
+    lazy val summary: ServiceCostEstimate.Summary =
+      forEnvironments
+        .foldLeft(Summary.zero)((summary, forEnv) => summary + Summary(forEnv.slots, forEnv.yearlyCostGbp))
```
colin-lamed, line 68: Since you've already defined `+` for Summary: `forEnviroments.map(_.toSummary).fold(Summary.zero)(_ + _)`
But we could skip the `toSummary` and use `_.values` if `forEnvironments` was modelled as a map

app/uk/gov/hmrc/cataloguefrontend/service/CostEstimationService.scala
```diff
@@ -58,21 +61,65 @@ object CostEstimationService {
...
+    final case class Config(slotCostPerYear: Double, totalAwsCostPerYear: String)
+
+    object Config {
+
+      val defaultSlotCostPerYear: Double = 650
+      val defaultTotalAwsCostPerYear: String = "£5.4M"
+
+      def get(): Config = {
+        val slotCostPerYear =
+          sys.props.get("cost-estimates.slot-cost-per-year")
```
colin-lamed, line 113: can we remove this `get` and inject the Config where needed? Then it can use configuration in config files as well as in system properties - and make testing easier etc..

app/uk/gov/hmrc/cataloguefrontend/service/CostEstimationService.scala
tomwadeson: Will do, there is indeed a monoidal `foldMap` waiting to get out! 🤓

app/uk/gov/hmrc/cataloguefrontend/service/CostEstimationService.scala
tomwadeson: I'll have a tinker, but suspect I'll only ever be operating on the `.values` projection if I were to do this. The one for each point is persuasive though 👍

tomwadeson: `.values` or `.toList`, that is

colin-lamed (approved):

PR #330, opened 2022-01-20
APPSEC-66 leak detection rules page

hmrc-phil:


app/uk/gov/hmrc/cataloguefrontend/connector/LeakDetectionConnector.scala
```diff
@@ -49,6 +50,20 @@ class LeakDetectionConnector @Inject() (
...
+    http
+      .GET[Seq[LeakDetectionRuleSummary]](
+        url"$url/api/rule-summaries",
+        headers = Seq("Accept" -> "application/json")
+      )
+      .recover {
+        case NonFatal(ex) =>
+          logger.error(s"An error occurred when connecting to $url: ${ex.getMessage}", ex)
+          Seq.empty
+      }
```
beny23, line 65: OBSERVATION: Does this mean we show a blank page if there's an error?

beny23 (dismissed):

app/uk/gov/hmrc/cataloguefrontend/connector/LeakDetectionConnector.scala
hmrc-phil: Yeah, we just get the page title.  I've removed this now so we get the standard error page instead.

beny23 (approved):

PR #332, opened 2022-01-25
BDOG-1573 

tomwadeson:
Introduces a new page to the Catalogue at `/service/:name/costs` (linked to from the service info page) that shows a graph of estimated costs over time:

cost-estimation-page

### DIFF
--- a/app/uk/gov/hmrc/cataloguefrontend/service/CostEstimationService.scala
+++ b/app/uk/gov/hmrc/cataloguefrontend/service/CostEstimationService.scala
@@ -149,10 +149,7 @@ object CostEstimationService {
     date: Instant,
     totalSlots: Int,
     yearlyCostGbp: Double
-  ) {
-    def +(other: CostedResourceUsageTotal): CostedResourceUsageTotal =
-      CostedResourceUsageTotal(date, totalSlots + other.totalSlots, yearlyCostGbp + other.yearlyCostGbp)
-  }
+  )
 
   object CostedResourceUsageTotal {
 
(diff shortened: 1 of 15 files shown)


tomwadeson:
I don't think the legend for the first graph, Yearly Cost (GBP), gives much. I'll look into hiding this.

app/uk/gov/hmrc/cataloguefrontend/service/CostEstimationService.scala
```diff
@@ -61,6 +91,110 @@ object CostEstimationService {
...
+    }
+  }
+
+  final case class CostedResourceUsageTotal(
+    date: Instant,
+    totalSlots: Int,
+    yearlyCostGbp: Double
+  ) {
+    def +(other: CostedResourceUsageTotal): CostedResourceUsageTotal =
+      CostedResourceUsageTotal(date, totalSlots + other.totalSlots, yearlyCostGbp + other.yearlyCostGbp)
```
colin-lamed, line 154: What if the dates are not the same? Should this bomb?Would it be better to split the Date out of the "monoid"? (A value object for slots/yearlyCost seems worthwhile?)

colin-lamed (dismissed):

app/uk/gov/hmrc/cataloguefrontend/service/CostEstimationService.scala
tomwadeson: Yeah, fair, I think I'll just move `+` to the local scope of where it's used (which is safe, as we group by `date` there and just smash them all together... semigroupally)

colin-lamed (approved):

PR #353, opened 2022-03-03
BDOG-1695 update Slack hyperlinks on team page to display slack chann…

Shnick:
…el name

### DIFF
--- a/app/views/TeamInfoPage.scala.html
+++ b/app/views/TeamInfoPage.scala.html
@@ -17,7 +17,7 @@
 @import _root_.views.html.partials.DependenciesTeamPartial
 @import uk.gov.hmrc.cataloguefrontend.DateHelper._
 @import uk.gov.hmrc.cataloguefrontend.{DisplayableTeamMember, TeamActivityDates, ViewMessages}
-@import uk.gov.hmrc.cataloguefrontend.connector.UserManagementConnector.{UMPError, TeamDetails}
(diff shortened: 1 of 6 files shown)


app/views/TeamInfoPage.scala.html
```diff
@@ -171,6 +167,29 @@ <h3 class="board__heading">Team Members</h3>
     @dependenciesTeamPartial(teamName, masterTeamDependencies, prodDependencies, showLink = true)
 }
 
+@addSlackInfo(label: String, slackInfo: SlackInfo) = {
+  @slackInfo match {
+    case SlackInfo(_, url) if !(url.startsWith("http://") || url.startsWith("https://"))  => {
```
colin-lamed, line 172: can we move these filters into `SlackInfo`?
Maybe `SlackInfo.hasValidUrl`.
Similarly `SlackInfo.hasValidName`?  (the regex could also be cached)

Could even model with something like `SlackInfo = SlackInfo(name, url) | InvalidUrl(name) | InvalidName(url)`?

colin-lamed (dismissed):

app/views/TeamInfoPage.scala.html
Shnick: Thanks, moved as suggested and created some unit tests.

colin-lamed (approved):

PR #363, opened 2022-04-04
BDOG-1713 Fix guest username in header menu

Shnick:


app/views/standard_layout.scala.html
```diff
@@ -120,7 +120,7 @@
                     <li class="dropdown">
                         @partials.with_username { username =>
                             <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">
-                                @username <span class="caret"></span>
+                                @if(username.startsWith("guest-")) { Guest } else { @username } <span class="caret"></span>
```
christopherjturner, line 123: would this be better done in the partial?

christopherjturner (approved):

app/views/standard_layout.scala.html
Shnick: Thanks changed.

colin-lamed (approved):

PR #365, opened 2022-04-05
BDOG-1712 adds search bar behind feature.search flag

christopherjturner:


app/uk/gov/hmrc/cataloguefrontend/search/IndexBuilder.scala
```diff
@@ -0,0 +1,83 @@
...
+}
+
+@Singleton
+class IndexBuilder @Inject()(teamsAndRepositoriesConnector: TeamsAndRepositoriesConnector)(implicit ec: ExecutionContext){
+
+  val cachedIndex = new AtomicReference[Seq[SearchTerm]]()
+
+  val hardcodedLinks = List(
+    new SearchTerm("dependency explorer",      catalogueRoutes.DependencyExplorerController.landing.url, "explorer") {
+      override lazy val term: String = "depex"
```
colin-lamed, line 44: this is potentially a list of search terms (space separated) which we just look for matches in? So should we call it `terms`?

app/uk/gov/hmrc/cataloguefrontend/search/IndexBuilder.scala
```diff
@@ -0,0 +1,83 @@
...
+    SearchTerm("lds",        leakRoutes.LeakDetectionController.ruleSummaries.url, "explorer"),
+  )
+
+  def buildIndexes(): Future[List[SearchTerm]] = {
+    implicit val hc = HeaderCarrier()
+    for {
+      repos         <- teamsAndRepositoriesConnector.allRepositories
+      teams         <- teamsAndRepositoriesConnector.allTeams
+      teamPageLinks =  teams.flatMap(t => List(SearchTerm(t.name.asString, teamRoutes.TeamsController.team(t.name).url, "teams"),
+                                               SearchTerm(t.name.asString, url"${wrwRoutes.WhatsRunningWhereController.releases(false).url}?profile_type=team&profile_name=${t.name.asString}".toString, "deployments")))
```
colin-lamed, line 59: should `profileType` and `profileName` be moved into the routes?

app/uk/gov/hmrc/cataloguefrontend/search/IndexBuilder.scala
```diff
@@ -0,0 +1,83 @@
...
+      repoLinks     =  repos.flatMap(r => List(SearchTerm(r.name, catalogueRoutes.CatalogueController.repository(r.name).url, "repo"),
+                                               SearchTerm(r.name, healthRoutes.HealthIndicatorsController.breakdownForRepo(r.name).url, "health"),
+                                               SearchTerm(r.name, leakRoutes.LeakDetectionController.report(r.name, r.defaultBranch).url,"leaks")))
+      serviceLinks  =  repos.filter(_.repoType == RepoType.Service).map(r => SearchTerm(r.name, catalogueRoutes.CatalogueController.serviceConfig(r.name).url, "config"))
+    } yield hardcodedLinks ++ teamPageLinks ++ repoLinks ++ serviceLinks
+  }
+
+  def getIndex(refresh: Boolean = false): Future[Seq[SearchTerm]] = {
```
colin-lamed, line 67: do we need the refresh flag? Thinking if we need a refresh, we just trigger the s`IndexScheduler` to run. That way we can guarantee we don't have it running twice at the same time etc.
Also since the scheduler runs after deployment (1 second) I don't think we need to worry about the cache being empty case either - if it's empty, it's more likely because the scheduler is already running?

app/uk/gov/hmrc/cataloguefrontend/search/SearchController.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  def search(query: String) =  Action.async { request =>
+    for {
+      index       <- indexBuilder.getIndex()
+      searchTerms  = query.split(" ", 2).filter(_.nonEmpty)
+      termFilter   = searchTerms.headOption.map(IndexBuilder.normalizeTerm).getOrElse("")
+      kindFilter   = searchTerms.tail.lastOption.map(_.toLowerCase)
+      matches      = index.filter(st => st.term.contains(termFilter))
+      matches2     = kindFilter.map(kf => matches.filter(_.linkType.toLowerCase.contains(kf))).getOrElse(matches)
+    } yield Ok(view(matches2.sortBy(_.term).take(20)))
```
colin-lamed, line 38: if we extract a `search(query: String): Seq[String]` function (leaving `Ok(view(matches))`) , we could add a test more easily to demonstrate how this works?

colin-lamed (approved):

colin-lamed (approved):

PR #380, opened 2022-05-06
BDOG-1582 Enable branch protection

tomwadeson:


app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
```diff
@@ -215,7 +222,8 @@ class CatalogueController @Inject() (
           envDatas                    = optLatestData.fold(envDatas)(envDatas + _),
           linkToLeakDetection         = urlIfLeaksFound,
           productionEnvironmentRoute  = serviceUrl,
-          serviceRoutes               = serviceRoutes
+          serviceRoutes               = serviceRoutes,
+          authorisation               = hasAuth
```
colin-lamed, line 226: both `authorisation` and `hasAuth` are vague in what auth we're talking about without looking at the type, and we could add more - wonder if they both should be called `hasBranchProtectionAuth`

app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
```diff
@@ -432,6 +474,29 @@ object RepoListFilter {
   )
 }
 
+object EnableBranchProtection {
+
+  sealed trait HasAuthorisation
+
+  object HasAuthorisation {
+
+    final case object Yes extends HasAuthorisation
```
colin-lamed, line 483: would a value object wrapping Boolean enough?

app/views/partials/branchProtection.scala.html
```diff
@@ -0,0 +1,85 @@
...
+        Enable
+    </button>
+  }
+  case _ => {
+    <button type="button"
+            class="btn btn-primary"
+            data-toggle="modal"
+            data-target="#enableBranchProtectionModal"
+            style="margin-left: 15px; margin-bottom: 15px"
+            disabled>
```
colin-lamed, line 57: can we just wrap the filter around the `disabled` property (less code, and easier to see the difference).
with a boolean wrapper it might be simpler 
```scala
@(!authorisation.asBoolean || repo.branchProtection.forall(!_.isProtected)) { disabled }
```
Is it worth not having the button at all if you don't have perms - might make it clearer to the user when they don't have the perms vs already enabled

colin-lamed (approved):

PR #392, opened 2022-05-25
BDOG-1877 Do filtering in service-dependencies Mongo instead of Catal…

jonnydh:
…ogue


app/uk/gov/hmrc/cataloguefrontend/BobbyRulesTrendController.scala
```diff
@@ -81,9 +81,8 @@ class BobbyRulesTrendController @Inject() (
                    hasErrors = formWithErrors => Future.successful(BadRequest(page(formWithErrors, allRules, flags = SlugInfoFlag.values, data = None))),
                    success = query =>
                      (for {
-                       violations <- EitherT.right[Result](serviceDeps.getHistoricBobbyRuleViolations)
+                       violations <- EitherT.right[Result](serviceDeps.getHistoricBobbyRuleViolations(query.rules.toList))
```
colin-lamed, line 84: you could just make `getHistoricBobbyRuleViolations` take a `Seq` - it doesn't look like it matters which you use here

colin-lamed (approved):

PR #394, opened 2022-06-08
BDOG-1883

Tyrpix:


app/views/partials/code_and_builds.scala.html
```diff
@@ -36,6 +37,9 @@ <h3 class="board__heading">Code and Build</h3>
             }
             <ul class="list list--minimal list-two-columns">
                 <li class="list-item"><a id="link-to-github" href="@{repo.githubUrl}" target="_blank" rel="noreferrer noopener">Github.com<span class="glyphicon glyphicon-new-window"/></a></li>
+                @if(repo.repoType.asString.equals("Service")) {
```
tomwadeson, line 40: We could use the `RepoType` type itself for this guard:

```
if (repo.repoType == RepoType.Service) { ...

File: app/views/whatsrunningwhere/DeploymentTimeline.scala.html [hunk 2]
```diff
@@ -88,10 +89,9 @@
                 return `<div class="timeline-tooltip"><b> ${ver} deployed by ${user} </b><br>From ${start} To  ${end}<br><em>Deployed for ${days} days</em> </div>`
             }
     </script>
-    <h2>Deployment Timeline @if(service != ""){ - @service}</h2>
+   <h2>Deployment Timeline @if(service != ""){<a id="link-to-service-info-page" href="@CatalogueController.repository(service)">- @service</a></h2>}
```
tomwadeson, line 92: I know it's not part of your change, but I wonder why we guard against the empty string here? If it's expected that this page could render without a given service, an `Option[String]` would be a better representation.

Either way, `CatalogueController.service` might make more sense than the generic `repository` endpoint too.

tomwadeson (commented):
Cool! A couple of comments

tomwadeson (approved):

PR #400, opened 2022-06-15
BDOG-1939 Further improvements to Catalogue Search

jonnydh:
1. Ensures that exact matches are always returned first, by increasing their weighting

2. Keeps SearchTerms belonging to the same service together by sorting on their name AFTER weighting.

### DIFF
--- a/app/uk/gov/hmrc/cataloguefrontend/search/SearchIndex.scala
+++ b/app/uk/gov/hmrc/cataloguefrontend/search/SearchIndex.scala
@@ -89,17 +89,20 @@ object SearchIndex {
   //       so we can reuse the partial result set
 
   private[search] def search(query: Seq[String], index: Map[String, Seq[SearchTerm]]): Seq[SearchTerm] = {
+    val ordering = Ordering.Tuple2(Ordering.Float.reverse, Ordering.String)  //Custom ordering to allow ascending & descending sort in same SortBy)
     val normalised = query.map(normalizeTerm)
     normalised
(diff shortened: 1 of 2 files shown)


app/uk/gov/hmrc/cataloguefrontend/search/SearchIndex.scala
```diff
@@ -89,17 +89,20 @@ object SearchIndex {
...
 
   private[search] def search(query: Seq[String], index: Map[String, Seq[SearchTerm]]): Seq[SearchTerm] = {
+    val ordering = Ordering.Tuple2(Ordering.Float.reverse, Ordering.String)  //Custom ordering to allow ascending & descending sort in same SortBy)
     val normalised = query.map(normalizeTerm)
     normalised
       .foldLeft(index.getOrElse(normalised.head.slice(0,3), Seq.empty)) {
         (acc, cur) => acc.filter(_.terms.exists(_.contains(cur)))
-      }.sortBy(_.weight)
+      }.map(st =>
+      if(normalised.exists(_.equalsIgnoreCase(st.name))) SearchTerm(st.linkType, st.name, st.link, 1f, st.hints) else st) //Increase weighting of an exact match
```
Shnick, line 98: Hey `copy` is really useful when updating case classes - https://alvinalexander.com/scala/scala-case-class-copy-method-example/
```
if (normalised.exists(_.equalsIgnoreCase(st.name))) st.copy(weight = 1f) else st)
```

Shnick (approved):

PR #405, opened 2022-06-23
Bdog 1947: Add instance & slot view to WhatsRunningWhere

jonnydh:


app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/WhatsRunningWhereController.scala
```diff
@@ -20,22 +20,22 @@ import cats.implicits._
...
 
 @Singleton
 class WhatsRunningWhereController @Inject() (
-  service           : WhatsRunningWhereService,
-  page              : WhatsRunningWherePage,
-  override val mcc  : MessagesControllerComponents,
-  override val auth : FrontendAuthComponents
+                                              service           : WhatsRunningWhereService,
+                                              page              : WhatsRunningWherePage,
+                                              instance_page     : WhatsRunningWhereInstances,
```
beny23, line 36: OBSERVATION: code style in scala here would be camel case (i.e. `instancePage`)

app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/WhatsRunningWhereController.scala
```diff
@@ -57,14 +57,32 @@ class WhatsRunningWhereController @Inject() (
...
-        profileNames = profiles.filter(_.profileType == selectedProfileType).map(_.profileName).sorted
+        environments          = distinctEnvironments(releases)
+        profileNames          = profiles.filter(_.profileType == selectedProfileType).map(_.profileName).sorted
       } yield Ok(page(environments, releases, selectedProfileType, profileNames, form, showDiff))
     }
+
+  def releasesConfig(showMemoryUse: Boolean): Action[AnyContent] = {
+    //The threshold of memory across instances and slots, for which the RGBA alpha value will be at its maximum.
+    //Any slotsAndInstancesToMemory values above this will be bounded to this figure.
+    val maxMemory = 32768.0
```
beny23, line 72: AMBER: I would read this from config and put the value into application.conf, so it can be overridden without a code change

Tyrpix: Added a class to the service, not sure it should have it's own config file just for a single function?

christopherjturner (approved):

PR #418, opened 2022-07-05
Bdog 1947 partial

jonnydh:


app/views/whatsrunningwhere/WhatsRunningWherePage.scala.html
```diff
@@ -135,6 +124,16 @@ <h1>What's Running Where?</h1>
...
         };
+
+        function selectViewMode(){
+            if(document.getElementById('view_mode').value === "versions") {
+                document.getElementById('view_mode').value = "instances"
+            } else {
+                document.getElementById('view_mode').value = "versions"
+            }
+            document.getElementById('form').submit();
+        }
```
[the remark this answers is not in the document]
jonnydh: I think so - Lines 83 and 85? :)

app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/WhatsRunningWhereController.scala
```diff
@@ -61,25 +62,13 @@ class WhatsRunningWhereController @Inject() (
         form                 <- Future.successful(WhatsRunningWhereFilter.form.bindFromRequest)
         profile               = profileFrom(form)
         selectedProfileType   = form.fold(_ => None, _.profileType).getOrElse(ProfileType.Team)
+        selectedViewMode      = form.fold(_ => None, _.viewMode).getOrElse(ViewMode.Versions)
         (releases, profiles) <- (service.releasesForProfile(profile).map(_.sortBy(_.applicationName.asString)), service.profiles).mapN((r, p) => (r, p))
         environments          = distinctEnvironments(releases)
+        serviceDeployments: Seq[ServiceDeploymentConfigSummary] <- service.allReleases(releases)
         profileNames          = profiles.filter(_.profileType == selectedProfileType).map(_.profileName).sorted
-      } yield Ok(page(environments, releases, selectedProfileType, profileNames, form, showDiff))
+      } yield Ok(page(environments, releases, selectedProfileType, profileNames, form, showDiff, serviceDeployments.sortBy(_.serviceName), config.maxMemoryAmount, selectedViewMode))
```
christopherjturner, line 70: wondering if the two controller approach might have been better here? you could avoid loading data you dont need to render a given view. If you wanted to keep a single endpoint/url you could break each view (including loading all its data) into its own function and have the endpoint call either one depending on the viewmode param?

hmrc-phil (dismissed):

app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/WhatsRunningWhereController.scala
hmrc-phil: Is it not just the `serviceDeployments`?  You could just wrap L68 with an `if else`

app/views/whatsrunningwhere/WhatsRunningWherePage.scala.html
```diff
@@ -74,45 +78,30 @@ <h1>What's Running Where?</h1>
...
-            </td>
-            </tr>
+            @if(viewMode.asString == "versions") {
+                <tbody class="list">
+                @WhatsRunningWhereVersions(whatsRunning, environments, showDiff)
+                </tbody>
+            } else {
+                <tbody class="list list-darker-text">
+                @WhatsRunningWhereInstances(serviceDeployments, environments, maxMemory)
+                </tbody>
```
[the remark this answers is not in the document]
jonnydh: Have addressed that in latest commit :)

app/uk/gov/hmrc/cataloguefrontend/whatsrunningwhere/WhatsRunningWhereController.scala
christopherjturner: i think you might need to do something about releasesForProfile when not looking at releases (theres a connector to get just profiles)

jonnydh: Have split created 2 functions that are called by the controller, meaning only the data we need is loaded :)

christopherjturner (approved):

PR #441, opened 2022-07-29
BDOG-2046 implement custom sorting function as List.js cannot handle …

jonnydh:
…negative numbers

### DIFF
--- a/app/views/HealthIndicatorsLeaderBoard.scala.html
+++ b/app/views/HealthIndicatorsLeaderBoard.scala.html
@@ -109,13 +109,7 @@
                 sortFunction: function(a, b) {
                     var a = parseInt(a._values.overallScore)
                     var b = parseInt(b._values.overallScore)
-                    if (a > b) {
-                        return 1;
-                    } else if (a < b) {
-                        return -1;
-                    } else {
-                        return 0;
-                    }
+                    return a - b;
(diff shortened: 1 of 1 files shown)


app/views/HealthIndicatorsLeaderBoard.scala.html
```diff
@@ -93,13 +93,33 @@ <h1>Health Indicator Leaderboard</h1>
...
 
+        //Add a custom sorting function, as the default List.js sorting algorithm cannot handle negative numbers.
+        //Note: A Listener is required, as the List.js custom 'sortFunction' only auto-fires on page load, not on button click.
+        document.getElementById("overall-score-button").addEventListener("click", function (e) {
+            indicatorsList.sort('overallScore', {
+                order: e.target.classList.contains("desc") ? "desc" : "asc",
+                sortFunction: function(a, b) {
+                    var a = parseInt(a._values.overallScore)
+                    var b = parseInt(b._values.overallScore)
+                    if (a > b) {
```
christopherjturner, line 112: does it have to be exactly in the range of -1 to 1? if not, i guess you could just return a - b for more or less the same effect?

christopherjturner (approved):

christopherjturner (approved):

PR #442, opened 2022-08-08
BDOG-2061: Changes button to drop down

Tyrpix:


app/views/whatsrunningwhere/WhatsRunningWherePage.scala.html
```diff
@@ -79,11 +79,10 @@ <h1>What's Running Where?</h1>
                 </div>
             </div>
             <input type="hidden" id="view_mode" name="view_mode" value="@viewMode.asString"/>
-            @if(viewMode.asString == "versions"){
-                <a style="float: right" onclick='selectViewMode("instances");'>instances</a>
-              } else {
-                <a style="float: right"  onclick='selectViewMode("versions");'>versions</a>
-              }
+            <select class="form-control" style="float: right; width: 120px;" onchange="selectViewMode(value)">
```
hmrc-phil, line 82: I don't think you'll need that function anymore - the only reason it exists is because we had multiple controls mapping to the same value.

christopherjturner (dismissed):

hmrc-phil (approved):

PR #444, opened 2022-08-11
BDOG-2077 - Host list.js locally, and modify list.js source code to allow search by punctuation

jonnydh:


christopherjturner (dismissed):

colin-lamed:
If you call the file `list-2.3.0-patched.min.js` it will be obvious we're not just self-hosting `list-2.3.0.min.js`

jonnydh:
> If you call the file `list-2.3.0-patched.min.js` it will be obvious we're not just self-hosting `list-2.3.0.min.js`

Done :) @colin-lamed

colin-lamed (approved):

PR #453, opened 2022-09-02
BDOG-1985 add pr-commenter recommendations page

Shnick:


app/uk/gov/hmrc/cataloguefrontend/connector/PrCommenterConnector.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  case class Report(
+    name     : String,
+    teamNames: List[String],
+    version  : Version,
+    comments : Seq[Comment],
+    created  : Instant
+  ) {
+    // Repos with joint ownership of 8+ teams, defines the repo as shared
+    def isShared: Boolean = teamNames.length >= 8
```
colin-lamed, line 41: move `8` to a constant (or config) now it's duplicated with `GitRepository`?

app/uk/gov/hmrc/cataloguefrontend/connector/PrCommenterConnector.scala
```diff
@@ -0,0 +1,87 @@
...
+class PrCommenterConnector @Inject()(
+  httpClientV2  : HttpClientV2,
+  servicesConfig: ServicesConfig
+)(implicit val ec: ExecutionContext) extends Logging {
+  import HttpReads.Implicits._
+  import PrCommenterConnector._
+
+  private val baseUrl = servicesConfig.baseUrl("pr-commenter")
+
+  implicit private val hc = HeaderCarrier()
```
colin-lamed, line 75: you should forward the caller's `HeaderCarrier` given this is not an async call

colin-lamed (approved):

app/uk/gov/hmrc/cataloguefrontend/connector/PrCommenterConnector.scala
```diff
@@ -38,7 +39,7 @@ object PrCommenterConnector {
     created  : Instant
   ) {
     // Repos with joint ownership of 8+ teams, defines the repo as shared
```
colin-lamed, line 41: probably don't need the comment (we shouldn't maintain the number in both comment and constant)

app/uk/gov/hmrc/cataloguefrontend/connector/PrCommenterConnector.scala
```diff
@@ -38,7 +39,7 @@ object PrCommenterConnector {
...
-    def isShared: Boolean = teamNames.length >= 8
+    def isShared: Boolean = teamNames.length >= Constant.sharedRepoTeamsCutOff
```
colin-lamed, line 42: Constant could be moved to `GitRepository` object - since that seems to own the value. We're just trying to keep the Report consistent.

app/uk/gov/hmrc/cataloguefrontend/connector/PrCommenterConnector.scala
```diff
@@ -38,7 +39,7 @@ object PrCommenterConnector {
     created  : Instant
```
colin-lamed, line 39: `Report` is used by the whole prcommenter feature (views, controllers, as well as connector) - I think adding a `prcommenter` feature package as we do for many other features makes sense - then we can move the model to that package.

colin-lamed (approved):

colin-lamed (approved):

PR #467, opened 2022-09-18
BDOG-2036 Fix CSS

jonnydh:
Currently only certain columns show the hand-pointer when hovering over, meaning only certain columns can expand the collapsable row. This is inconsistent. Desired behaviour is that clicking anywhere on the parent row should expand the child row.


jonnydh:
> * addSbtPlugin("uk.gov.hmrc" % "sbt-auto-build" % "3.8.0")

Hey PR Commenter, it's already there :O

RikSherman (approved):

PR #537, opened 2023-02-16
BDOG-2362 Libraries - Show all versions dependencies

RikSherman:


christopherjturner (approved):

app/views/LibraryInfoPage.scala.html
```diff
@@ -29,7 +29,7 @@
 )
 
 @(library                : GitRepository,
-  repositoryModules      : Option[RepositoryModules],
+  metaInfo               : Seq[RepositoryModules],
```
colin-lamed, line 32: The name `metaInfo` is a bit ambiguous (especially when you see things like `metaInfo.headOption`), I'd be tempted to leave it as `repositoryModules` (even if there is a lack of double plural)?

app/views/LibraryInfoPage.scala.html
```diff
@@ -51,7 +51,7 @@ <h1>
   </header>
 
   @partials.leak_detection_banner(linkToLeakDetection)
-  @defining(repositoryModules.fold(Seq.empty[Dependency])(_.allDependencies)){ dependencies =>
+  @defining(metaInfo.headOption.fold(Seq.empty[Dependency])(_.allDependencies)){ dependencies =>
```
colin-lamed, line 54: can we use `@defining(metaInfo.headOption){ latestVersion =>` to explain what's going on, and avoid multiple calls to `metaInfo.headOption`?

app/views/LibraryInfoPage.scala.html
```diff
@@ -78,10 +78,39 @@ <h1>
       <div class="col-md-6">
         @partials.build(library, hasBranchProtectionAuth, commenterReport)
       </div>
+
+        <div class="col-md-6">
+        @partials.dependency_legend(true)
```
colin-lamed, line 83: what is `true`? can we use a named param?

app/views/partials/ModulesPartial.scala.html
```diff
@@ -21,7 +21,9 @@
 )
 
 @(
-  repositoryModules: RepositoryModules
+  repositoryModules : RepositoryModules,
+  isLatest          : Boolean         = false,
```
colin-lamed, line 25: wonder if naming `isLatest` as `displayBuildInfo` will decouple it from domain knowledge, and make it easier to understand why it's being asked for?

PR #565, opened 2023-04-14
BDOG-1992 filter config display by deployed env

peteanning:
Displays configs always for local and by deployed environment for all other envs

### DIFF
--- a/app/views/ServiceConfigPage.scala.html
+++ b/app/views/ServiceConfigPage.scala.html
@@ -18,11 +18,13 @@
 @import uk.gov.hmrc.cataloguefrontend.service.ConfigService
 @import uk.gov.hmrc.cataloguefrontend.service.ConfigService.{ConfigByKey, ConfigEnvironment, ConfigSourceValue}
 @import scala.collection.immutable.ListMap
+@import uk.gov.hmrc.cataloguefrontend.whatsrunningwhere.WhatsRunningWhereVersion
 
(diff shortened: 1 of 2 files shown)


app/views/ServiceConfigPage.scala.html
```diff
@@ -46,7 +48,7 @@ <h1 id="service-header">Config: @serviceName</h1>
                     <div id="header-0" class="grid-item sticky-column truncate-content" style="width:300px; text-indent: 1.2em">Key</div>
                     <div class="gutter gutter-horizontal"></div>
                     @for((env, index) <- ConfigEnvironment.values.zipWithIndex) {
-                        <div id="header-@{index + 1}" class="grid-item truncate-content column-@env.asString" style="width:500px;">@env.asString</div>
+                        <div id="header-@{index + 1}" class="grid-item truncate-content column-@env.asString @if(!deployments.exists(_.environment.asString == env.asString) && env.asString != "local") {hidden} else{}" style="width:500px;">@env.asString</div>
```
colin-lamed, line 51: If the types are the same, you could do `deployments.exists(_.environment == env)`
I don't think the `else {}` is needed.

colin-lamed (approved):

app/views/ServiceConfigPage.scala.html
peteanning: One is an Environment and the other is a ConfigEnvironment

PR #566, opened 2023-04-17
Bdog 1992

peteanning:
Fixed issue with only hiding config column headers

### DIFF
--- a/app/views/ServiceConfigPage.scala.html
+++ b/app/views/ServiceConfigPage.scala.html
@@ -30,7 +30,7 @@
 )
 
 @hasDeployment(env: ConfigEnvironment, textIfTrue: String, textIfFalse: String) = @{
-    if ((deployments.map(_.environment.asString) :+ "local").contains(env.asString)) {
+    if ((deployments.map(e => ConfigEnvironment.ForEnvironment.apply(e.environment)) :+ ConfigEnvironment.Local ).contains(env)) {
        textIfTrue
     } else {
       textIfFalse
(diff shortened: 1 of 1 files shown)


app/views/ServiceConfigPage.scala.html
```diff
@@ -29,9 +29,11 @@
   request: RequestHeader
 )
 
-@display(env: String, textIfTrue: String, textIfFalse: String = "") = {
-    @defining(deployments.map(_.environment.asString) :+ "local"){deployedEnvs =>
-        @if(!deployedEnvs.contains(env)){ @textIfTrue } else { @textIfFalse }
+@hasDeployment(env: ConfigEnvironment, textIfTrue: String, textIfFalse: String = "") = @{
+    if ((deployments.map(_.environment.asString) :+ "local").contains(env.asString)) {
```
colin-lamed, line 33: Could avoid comparing the `asString` representation with `(deployments.map(ConfigEnvironment.ForEnvironment.apply) :+ ConfigEnvironment.Local).contains(env)`

colin-lamed: Providing both `textIfTrue` and `textIfFalse` runs the risk of being mixed up, at least harder to follow (note, it's being called without named params).
I think using `@defining(..){deployments =>` which would move the boolean logic to the call points could be simpler? e.g. `@if(deployments.contains(env) { "hidden"}`

peteanning: Added named parameters as discussed

peteanning: Refactored to compare types not Strings

colin-lamed (approved):

PR #581, opened 2023-05-05
BDOG-2530 shared teams fragment

fionahiklas:
Implementing fixes for improvement 3 from https://jira.tools.tax.service.gov.uk/browse/BDOG-2530


README.md
```diff
@@ -11,6 +11,20 @@
 
 See [catalogue-acceptance-tests]("https://github.com/hmrc/catalogue-acceptance-tests") for help setting up dependent services.
 
+Run docker with the following command
```
colin-lamed, line 14: This is already documented in the above `catalogue-acceptance-tests`, maybe could just reword "for help setting up dependent services and mongo"

app/uk/gov/hmrc/cataloguefrontend/connector/TeamsAndRepositoriesConnector.scala
```diff
@@ -173,9 +173,6 @@ case class GitRepository(
 ) {
   def isShared: Boolean =
```
colin-lamed, line 174: `isShared` is only used by the partial - it could be moved into the partial, along with the `Constant.sharedRepoTeamsCutOff`.

app/views/partials/GitRepositoryTeamNames.scala.html
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ *@
+
+@import uk.gov.hmrc.cataloguefrontend.connector.GitRepository
+@import uk.gov.hmrc.cataloguefrontend.connector.model.TeamName
+
+@this()
+
+@(
+        repository: GitRepository
```
colin-lamed, line 23: We have two partials that do the same, to support different input types.
Can we change this to take `Seq[Team]` rather than `GitRepository` or `ServiceWithDependency`, then we only need one.
The `isShared` function can be inlined here. Inlining the `Constant.sharedRepoTeamsCutOff` will make it clear that the cutoff is `8`, but the `teamsNames.take` is `15` - which don't align. (`splitAt(8)` and checking if the second list is empty might avoid defining the value more than once)

app/views/partials/GitRepositoryTeamNames.scala.html
```diff
@@ -0,0 +1,32 @@
...
+)
+
+@if(!repository.isShared) {
+    @repository.teamNames.map {tn => <div  class="repo-team" ><a href="@uk.gov.hmrc.cataloguefrontend.teams.routes.TeamsController.team(TeamName(tn))">@tn</a></div> }
+} else {
+    <div title="@repository.teamNames.take(15).mkString("\n")" class="repo-team" >
+        <em><a href="@uk.gov.hmrc.cataloguefrontend.routes.CatalogueController.repository(repository.name)#teams">Shared by @repository.teamNames.length teams</a></em>
```
colin-lamed, line 30: I think we can leave this link out. It might be unexpected to the client to be taken to a different page when clicking on `Shared by ...` - it stands out more when not a link. There's usually a link already available on the repository name in the same table.

test/uk/gov/hmrc/cataloguefrontend/JsonData.scala
```diff
@@ -22,78 +22,21 @@ import uk.gov.hmrc.cataloguefrontend.shuttering.{ShutterStatusValue, ShutterType
 import java.time.Instant
 
 /**
-  * Created by armin.
-  */
+ * Created by armin.
+ *
+ * TODO: Refactor the data here into files under the
```
colin-lamed, line 27: Can these TODO comments be removed now?
You can also remove the "Created by" comment while we're here.

app/uk/gov/hmrc/cataloguefrontend/connector/TeamsAndRepositoriesConnector.scala
colin-lamed: Actually, it looks like `prcommenter/model.scala` also duplicates this - can this also use the new partial?

README.md
fionahiklas: I think this might be backwards: we should have some details about how to run/test the catalogue-frontend locally in this README rather than in the acceptance tests.  Maybe a wider discussion for later about making development easier :)

app/views/partials/GitRepositoryTeamNames.scala.html
fionahiklas: The link is deliberate to make the table accessible and consistent.  For example, if someone uses a reader that zooms into part of the screen where the teams column is visible it's not ideal that they then have to scroll sideways to get to the link.

README.md
fionahiklas: Removed duplicate docs

app/uk/gov/hmrc/cataloguefrontend/connector/TeamsAndRepositoriesConnector.scala
fionahiklas: Updated `prcommenter/model.scala` and added tests for the PrCommenter page

app/views/partials/GitRepositoryTeamNames.scala.html
fionahiklas: The `GitRepository` object still needs it's `isShared` function and it's used in quite alot of code.  I can raise another ticket to handle this OR we can do it under this ticket?

README.md
```diff
@@ -3,13 +3,19 @@
...
 * [License](#license)
 
 ## Setup
 
-See [catalogue-acceptance-tests]("https://github.com/hmrc/catalogue-acceptance-tests") for help setting up dependent services.
+For help setting up dependent services see [catalogue-acceptance-tests]("https://github.com/hmrc/catalogue-acceptance-tests")
+
+## Development
+
+In order to run unit tests against any changes to views or partials you need to execute
```
colin-lamed, line 17: `sbt test` already invokes `compile` - did you mean `sbt clean`?

app/uk/gov/hmrc/cataloguefrontend/connector/TeamsAndRepositoriesConnector.scala
colin-lamed: Can `isShared` be removed from here now?

app/views/partials/TeamNames.scala.html
```diff
@@ -0,0 +1,34 @@
...
+
+@this()
+
+@(
+        thingLinkFunction: (String) => Call,
+        thingName: String,
+        teamNames: Seq[Any]
+)
+
+@if(teamNames.length < Constant.sharedRepoTeamsCutOff ) {
```
colin-lamed, line 28: If the constant is only required in this file, then it could be inlined here.

app/views/partials/TeamNames.scala.html
```diff
@@ -0,0 +1,34 @@
...
+ *@
+
+@import uk.gov.hmrc.cataloguefrontend.connector.model.TeamName
+@import uk.gov.hmrc.cataloguefrontend.config.Constant
+
+@this()
+
+@(
+        thingLinkFunction: (String) => Call,
+        thingName: String,
```
colin-lamed, line 24: A better name would be `repositoryName`

app/views/partials/TeamNames.scala.html
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ *@
+
+@import uk.gov.hmrc.cataloguefrontend.connector.model.TeamName
+@import uk.gov.hmrc.cataloguefrontend.config.Constant
+
+@this()
+
+@(
+        thingLinkFunction: (String) => Call,
```
colin-lamed, line 23: It looks like this param exists to support both `repository` and `service` urls. I think using the `repository` endpoint for all cases should be fine, so the reverse route will not need to be passed as a param. (service endpoint only exists for the edge case where the service name doesn't match the repo name, but this is rare, so don't think it's worth over engineering here)

test/uk/gov/hmrc/cataloguefrontend/JsonData.scala
colin-lamed: Or just keep the TODOs locally

app/views/partials/GitRepositoryTeamNames.scala.html
colin-lamed: OK - we can leave `isShared`

README.md
fionahiklas: Sorry, I should have clarified that comment - "run unit tests in the IDE" ... maybe there is an IDE setting but OOTB running a unit test (right click on it in the IDE) should run the required steps for converting partials to scala and then compiling that.  I couldn't see a task that did just the work of partial -> scala so the only solution seemed to be compile.  That seemed to work and appeared to be incremental anyway so not too time consuming.  Maybe there is a way to get the IDE to do this?  Or maybe the advice should be in the README so we don't have to get into IDE config or how people develop code

test/uk/gov/hmrc/cataloguefrontend/JsonData.scala
fionahiklas: The point for the TODOs was because they need to be TOdone :D And it might not always be me TODO'ing them so a local copy wouldn't really help much.  I've added https://jira.tools.tax.service.gov.uk/browse/BDOG-2593 so that the work to tidy up isn't lost

app/views/partials/TeamNames.scala.html
fionahiklas: I've left it parameterized because the table in each page has columns that go to different pages (repository or service), from a usability perspective we should probably stay consistent rather than have the link go to the repository in the case where other links go to service.

I guess if we refactor the code in the future so that the service page/link doesn't exist anymore then we can probably drop the distinction and the parameter.

app/uk/gov/hmrc/cataloguefrontend/connector/model/model.scala
```diff
@@ -23,7 +23,9 @@ import play.api.mvc.{PathBindable, QueryStringBindable}
 
 case class Username(value: String) extends AnyVal
 
-case class TeamName(asString: String) extends AnyVal
+case class TeamName(asString: String) extends AnyVal {
+  override def toString: String = asString
```
colin-lamed, line 27: Can we keep `toString` as the default. For logging, debugging, reading test failures etc. it's useful to see the type. This also aligns with all our other Stringable value classes.

app/views/partials/TeamNames.scala.html
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ *@
+
+@import uk.gov.hmrc.cataloguefrontend.connector.model.TeamName
+@import uk.gov.hmrc.cataloguefrontend.config.Constant
+
+@this()
+
+@(
+        controllerLinkString: String,
```
colin-lamed, line 23: The parameter name doesn't need to include the type (that info is already available). The name doesn't make it clear what it's used for - I'd suggest `sharedTeamsLink`

app/views/partials/TeamNames.scala.html
```diff
@@ -0,0 +1,33 @@
...
+        teamNames: Seq[Any]
```
colin-lamed, line 24: Can the parameter order be swapped? `teamNames` is the primary param, the rest being auxiliary (`sharedTeamsLink` could even have been optional with a default value of `None`)

app/views/partials/TeamNames.scala.html
```diff
@@ -0,0 +1,33 @@
...
+)
+
+@if(teamNames.length < Constant.sharedRepoTeamsCutOff ) {
+    @teamNames.map {tn => <div  class="repo-team" ><a href="@uk.gov.hmrc.cataloguefrontend.teams.routes.TeamsController.team(TeamName(tn.toString()))">@tn.toString()</a></div> }
+} else {
+    <div title="@teamNames.take(Constant.sharedRepoTeamsCutOff).mkString("\n")" class="repo-team" >
+        <em><a href="@controllerLinkString">Shared by @teamNames.length teams</a></em>
+    </div>
+}
```
colin-lamed, line 33: IDE can be configured to always add `\n` at EOF

test/uk/gov/hmrc/cataloguefrontend/JsonData.scala
colin-lamed: Can they be removed now we have a ticket?

colin-lamed (approved):

app/views/partials/TeamNames.scala.html
fionahiklas: Fixed

app/views/partials/TeamNames.scala.html
fionahiklas: Fixed

test/uk/gov/hmrc/cataloguefrontend/JsonData.scala
fionahiklas: I guess the point they can be removed is when they are TOdone :)

PR #604, opened 2023-06-09
BDOG-2616 - Add initial Create a Repository page in Catalogue

jonnydh:
- Added link on Repositories catalogue page
- Created view, following style of Jenkins jobs. Note that a deliberate decision was made NOT to use the Play field form helpers in the view template. This is because in order for information to flow semantically correct on the form, for each field, we would want the label (e.g. repository type), extra information, and then the form field itself. However if you use the play helpers, they will always construct a label (even if not specified), meaning that the label will always be directly before the input field, meaning any extra information for that field would have to be before the fields heading label, or after the input itself. 
- Created controller and form. Form uses the Play verifying/contraints helpers. 
- Created an Enum for CreateARepositoryType for easier validation


test/uk/gov/hmrc/cataloguefrontend/createarepository/CreateRepoFormSpec.scala
```diff
@@ -0,0 +1,256 @@
...
+  "repoNameLengthValidation" when {
+    "the repo name contains less than 48 characters" should {
+      "return false" in {
+        CreateRepoForm.repoNameLengthValidation("helloworld") shouldBe true
+      }
+    }
+
+    "the repo name contains exactly 47 characters" should {
+      "return true" in {
+        CreateRepoForm.repoNameLengthValidation("helloworldhelloworldhelloworldhelloworldhellowo") shouldBe true
```
colin-lamed, line 68: You could do something like `LazyList.continually("a").take(47).mkString` to avoid handcrafting strings of particular lengths

colin-lamed (approved):

PR #612, opened 2023-06-19
BDOG-2511 prevent reset password if prototype status is not Running

jordanrowe:


app/uk/gov/hmrc/cataloguefrontend/CatalogueController.scala
```diff
@@ -313,7 +314,8 @@ class CatalogueController @Inject() (
         hasPasswordChangeAuth,
         form,
         successMessage,
-        commenterReport
+        commenterReport,
+        prototypeStatus.status
```
colin-lamed, line 318: if we're only interested in the actual status (`PrototypeStatus`), should we encapsulate the temporary `GetPrototypeStatusResponse` in the connector - i.e. the connector should verify the `success` status if necessary

colin-lamed (approved):

PR #712, opened 2023-10-09
BDOG-2828 filter what's running where by version discrepancies betwee…

jordanrowe:
…n environments


test/uk/gov/hmrc/cataloguefrontend/model/EnvironmentQueryStringBindableSpec.scala
```diff
@@ -39,19 +39,19 @@ class EnvironmentQueryStringBindableSpec extends AnyWordSpec with Matchers with
     "fail to be bound when there is no associated value" in {
       val params = Map("environment" -> Seq.empty)
 
-      Environment.queryStringBindable.bind(key = "env", params).value shouldBe 'Left
```
victorarbuesmallada, line 42: can be ignored but we could compare with the expected left value?
```
    "fail to be bound when there is no associated value" in {
      val params = Map("environment" -> Seq.empty)

      Environment.queryStringBindable.bind(key = "env", params).value shouldBe Left("missing environment value")
    }

    "fail to be bound when there is more than one associated value" in {
      val params = Map("environment" -> Seq(Environment.Production.asString, Environment.QA.asString))

      Environment.queryStringBindable.bind(key = "env", params).value shouldBe Left("too many environment values")
    }

    "fail to be bound when the value is unrecognised" in {
      val params = Map("environment" -> Seq("unknown"))

      Environment.queryStringBindable.bind(key = "env", params).value shouldBe Left("Invalid Environment 'unknown'")
 ```

victorarbuesmallada (approved):

PR #722, opened 2023-11-01
BDOG-2681: Increase non performant query links time range

victorarbuesmallada:
Change layout of telemetry links


victorarbuesmallada:
From
Screenshot from 2023-11-01 17-24-31
To
Screenshot from 2023-11-01 17-24-18

jordanrowe (approved):

PR #797, opened 2024-01-12
BDOG-2931: Estimated cost explorer page

MatthewAlexanderBeck:


app/uk/gov/hmrc/cataloguefrontend/costs/CostsSummaryController.scala
```diff
@@ -0,0 +1,99 @@
...
+    Form(
+      mapping(
+        "team" -> optional(text).transform[Option[TeamName]](_.filter(_.trim.nonEmpty).map(TeamName.apply), _.map(_.asString))
+      )(RepoListFilter.apply)(RepoListFilter.unapply)
+    )
+}
+
+case class ServiceAndEnvironment(serviceName: String, configs: Seq[DeploymentConfig]) {
+
+  val getIntegrationEnvironment: Option[DeploymentConfig]   = configs.find(_.environment == Integration)
```
colin-lamed, line 73: The name suggests it returns an environment?
You just need a single `deploymentConfigs(env: Environment)` function?

app/views/costs/CostsSummaryPage.scala.html
```diff
@@ -0,0 +1,157 @@
...
+                    <th>Instances</th>
+                </tr>
+            </thead>
+            <tbody class="list" id="matches">
+            @configs.zipWithIndex.map { case (config, index) =>
+            <tr>
+                <td class="repo-name" id="row@{
+                    index
+                }_name"><a id="link-to-@config.serviceName-page" href="/repositories/@config.serviceName">@config.serviceName</a></td>
+                <td style="text-align: center">@config.getIntegrationEnvironment.map(_.deploymentSize.slots)</td>    @* Integration - slots *@
```
colin-lamed, line 106: We should iterate over `Environment.values` - this already has a predefined order (examples on other pages with env columns). It would reduce the code - and the use of a single extractor for all envs even more useful. `@config.deploymentConfigs(env).map(_.deploymentSize.slots)`

Since we're only ever interested in deploymentSize - the extractor could even be `@config.deploymentSize(env)`.
Infact, does `ServiceAndEnvironment` even need to keep `Seq[DeploymentConfig]`? Maybe just `Map[Environment, DeploymentSize]`

The comment seems unnecessary.

app/views/costs/CostsSummaryPage.scala.html
```diff
@@ -0,0 +1,157 @@
...
+@import uk.gov.hmrc.cataloguefrontend.ViewMessages
+@import uk.gov.hmrc.cataloguefrontend.connector.Team
+@import uk.gov.hmrc.cataloguefrontend.costs.ServiceAndEnvironment
+@import views.html.helper.select
+@import uk.gov.hmrc.cataloguefrontend.service.CostEstimateConfig
+
+@this(viewMessages: ViewMessages)
+
+@(
+        configs             : Seq[ServiceAndEnvironment],
```
colin-lamed, line 26: the naming could be clearer - `ServiceAndEnvironment` is called `configs` - and rather than representing a `ServiceName` and `Environment`. It seems to be (practically) `Map[ServiceName, Map[Environment, DeploymentSize]`. It might be better modelled as such (which is clear) - extractors being recognised provided ones (`get`) (and a couple of helpers if necessary).

conf/app.routes
```diff
@@ -81,6 +81,8 @@ GET        /config/search/results                       uk.gov.hmrc.cataloguefro
 GET        /config/warnings/search                      uk.gov.hmrc.cataloguefrontend.serviceconfigs.ServiceConfigsController.configWarningLanding()
 GET        /config/warnings/search/results              uk.gov.hmrc.cataloguefrontend.serviceconfigs.ServiceConfigsController.configWarningResults()
 
+GET        /estimated-cost-explorer                     uk.gov.hmrc.cataloguefrontend.costs.CostsSummaryController.onPageLoad(team: Option[String] ?= None)
```
colin-lamed, line 84: Do we need `estimated` in the URL?  It doesn't look like we're disambiguating with another cost-explorer - so I'd be tempted to leave this to an explanation on the page itself.

colin-lamed: Should the controller be called `CostExplorerController` - if that's how we're thinking of it it would make finding it easier

app/uk/gov/hmrc/cataloguefrontend/costs/CostsSummaryController.scala
```diff
@@ -0,0 +1,99 @@
...
+  costsSummaryPage: CostsSummaryPage,
+  costEstimateConfig: CostEstimateConfig,
+  override val mcc: MessagesControllerComponents,
+  override val auth: FrontendAuthComponents
+)(implicit
+  override val ec: ExecutionContext
+) extends FrontendController(mcc)
+    with CatalogueAuthBuilders {
+
+  def onPageLoad(team: Option[String] = None): Action[AnyContent] =
```
colin-lamed, line 49: can we call this `landing`? - `onPageLoad` sounds like event handling - `landing` is the naming used elsewhere.
If this only has one page, could just call it `costExplorer`

colin-lamed (approved):

colin-lamed (approved):

PR #843, opened 2024-03-04
BDOG-2168: Removed prototype as dependency explorer option

MatthewAlexanderBeck:


app/uk/gov/hmrc/cataloguefrontend/connector/TeamsAndRepositoriesConnector.scala
```diff
@@ -42,6 +42,8 @@ object RepoType {
 
   val values: List[RepoType] = List(Service, Library, Prototype, Test, Other)
 
+  val dependencyExplorerValues: List[RepoType] = List(Service, Library, Test, Other)
```
colin-lamed, line 45: dependency explorer doesn't need to pollute the `RepoType` model. Just add `RepoType.values.remove(RepoType.Prototype)` to DependencyExplorer, which will make it clear what it supports too

BriWak (approved):

colin-lamed (approved):

PR #850, opened 2024-03-07
BDOG-3013: Cost explorer CSV button

MatthewAlexanderBeck:


jordanrowe (approved):

app/uk/gov/hmrc/cataloguefrontend/costs/CostExplorerController.scala
```diff
@@ -43,13 +51,53 @@ class CostsSummaryController @Inject() (
...
+              s"instances.{${env.asString}}" -> configList.find(_.environment == env).map(_.deploymentSize.instances.toString).getOrElse("")
+            )
+        }
+
+        Seq("Application Name" -> serviceName) ++ slotsAndInstances ++ Seq("Estimated Cost (£ / year)" -> configEstimateCosts)
+    }.toSeq
+  }
 
+  def costExplorer(
+    name: Option[String] = None,
```
colin-lamed, line 73: do we need the `name` filter? It's misleading since it only works if `asCSV` is true and CSVs can be filtered as required once exported.

app/uk/gov/hmrc/cataloguefrontend/costs/CostExplorerController.scala
```diff
@@ -43,13 +51,53 @@ class CostsSummaryController @Inject() (
...
+    team: Option[String] = None,
+    asCSV: Option[Boolean] = None
```
colin-lamed, line 75: This could have type `Boolean` - the routes can define it is `false` if not present

colin-lamed (approved):

PR #878, opened 2024-04-17
BDOG-3096: Service deprecated banners and alerts

MatthewAlexanderBeck:


app/uk/gov/hmrc/cataloguefrontend/serviceconfigs/ServiceConfigsService.scala
```diff
@@ -352,6 +352,10 @@ object ServiceConfigsService {
     def hasDecommissioningDownstream: Boolean =
```
colin-lamed, line 352: can we remove this now?

app/views/partials/details.scala.html
```diff
@@ -121,7 +121,22 @@
...
+                    @lifecycle match {
+                        case Lifecycle(Deprecated, _, _) => {
+                            <label class="fw-bold">Status <a href="#"
+                            tabindex="0"
+                            aria-label="Status"
+                            data-bs-toggle="popover"
+                            data-bs-trigger="focus"
+                            data-bs-placement="left"
+                            data-bs-html="true"
+                            data-bs-content='Deprecated status is defined by the <a href="@repositoryDetails.githubUrl/blob/main/repository.yaml">repository.yaml</a>. To change this you will need to either remove the `deprecated` key or change its value to `false`. Changes can take up to 30 minutes to appear in the catalogue.'>
```
colin-lamed, line 133: Wonder if we should be vague and just say it can take a few minutes? This number will not be maintained as we change scheduler configuration etc. (It should be pretty quick on a webhook, else it can take a long time for a scheduler to catch it)

app/views/partials/service_relationships.scala.html
```diff
@@ -81,3 +82,10 @@
     }
     </ul>
 }
+
+@popoverContent(lifecycleStatus: Option[LifecycleStatus]) = {
+    @lifecycleStatus match {
+      case Some(Deprecated) => { "Marked as deprecated" }
```
colin-lamed, line 88: Should we just say "Deprecated"? The "Marked" terminology was introduced for decommissioning to show that it's still in the process and not decommissioned just yet.

app/views/ServiceInfoPage.scala.html
```diff
@@ -80,14 +80,23 @@ <h1 id="service-header" class="page-heading mt-4 position-relative">
       @partials.bobby_violations_banner(environment = Some(Environment.Production), dependencies = dependencies, pending = false)
       @partials.bobby_violations_banner(environment = Some(Environment.Production), dependencies = dependencies, pending = true )
     }
-    @if(serviceRelationships.hasDecommissioningDownstream) {
+
+    @if(serviceRelationships.hasDeprecatedDownstream) {
       <div class="alert alert-warning">
           <strong>
-              One or more of your <a id="service-relationships-anchor-link" href="#service_relationships">downstream dependencies</a> has been marked for decommissioning.
+              One or more of your <a id="service-relationships-deprecated-anchor-link" href="#service_relationships">downstream dependencies</a> has been marked as deprecated.
```
colin-lamed, line 87: Can we just say "has been deprecated"?

colin-lamed (approved):

app/uk/gov/hmrc/cataloguefrontend/serviceconfigs/ServiceConfigsService.scala
MatthewAlexanderBeck: I opted to keep in the banner for decomissioned downstream services (using this def). I was thinking though users should have made sure that no downstream services are using this before they get to the decomission step it may not stop them from disregarding the instructions. Thought keeping this in may help that just in case rather than possibly losing sight of these services.

colin-lamed: It's easily added back if we want to do this. And dead-code will just be deleted by someone else (unless intention is commented).

I expect if we wanted to do this, we'd just extend `hasDeprecatedDownstream` to cover both states for simplicity (Note Decommissioning status is really only intended for the service owners to do their job and is expected to be a short-lived status - Deprecation status is the one to advise for clients).

I think our efforts would be best spent on highlighting archived/deleted downstream dependencies if we end up in this state. It could encourage config clean up if nothing else.

colin-lamed (approved):

colin-lamed (approved):

